PR #6, opened 2017-02-10
correct spelling mistake

EdwardBetts:


stevegrubb:
The cut over to git is not complete just yet. (Will be switching by the end of the month.) I applied this manually to the svn repo. Thanks for reporting this.

PR #20, opened 2017-05-05
Add config option to control how many times audispd can write to syslog.

x746e:
I'm considering adding monitoring for audispd queue overflows by,
essentially, grepping /var/log/messages.  Currently audispd will only
log up to 5 (QUEUE_FULL_LIMIT) messages before stopping writing anything
to the log [1].  That makes the logging significantly less useful.

[1] https://github.com/linux-audit/audit-userspace/blob/master/audisp/queue.c#L96

The change adds a `overflow_warning_limit` option to audispd.conf to
control this.  It also can be set to zero to disable the limit
completely.


stevegrubb:
Closing this out since audispd no longer exists.

PR #21, opened 2017-05-07
errormsg: add descriptive macros to replace overloaded error codes

rgbriggs:
Several return codes were overloaded and no longer giving helpful error
return messages from the field and comparison functions
audit_rule_fieldpair_data() and audit_rule_interfield_comp_data().

Introduce 2 new macros with more helpful error descriptions for filter
missing, incompatible comparison data.

See: https://github.com/linux-audit/audit-userspace/issues/12

Signed-off-by: Richard Guy Briggs <rgb@redhat.com>

PR #22, opened 2017-05-07
errormsg table macro conversion

rgbriggs:
This should be applied after https://github.com/linux-audit/audit-userspace/pull/21 has been applied and the earlier patch should then be ignored by git since it is already applied.

### DIFF
--- a/lib/errormsg.h
+++ b/lib/errormsg.h
@@ -35,41 +35,6 @@ struct msg_tab {
 };
 
 #ifndef NO_TABLES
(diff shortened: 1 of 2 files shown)


rgbriggs:
Patch posted upstream:
 https://www.redhat.com/archives/linux-audit/2017-June/msg00047.html
Patch accepted upstream:
 https://github.com/linux-audit/audit-userspace/commit/b389789acf7ef70c6f28d4ec3381f57400d4acef

PR #23, opened 2017-05-11
Config paths

daniel-j-born:
Add command line options to auditd and audispd for config file paths.

### DIFF
--- a/audisp/audispd.c
+++ b/audisp/audispd.c
@@ -34,6 +34,7 @@
 #include <sys/poll.h>
 #include <netdb.h>
 #include <arpa/inet.h>
+#include <getopt.h>
 
 #include "audispd-config.h"
 #include "audispd-pconfig.h"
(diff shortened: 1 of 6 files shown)


stevegrubb:
I am interested in pulling this in, but I would have done some things differently. Do you have this as 1 patch rather than the 5 commits? If so, I'd take it via email. I'm readying a 2.8 release in the next couple of weeks. This would go into the new release.

daniel-j-born:
I rebase into a single commit. You can get a patch from github by putting ".patch" at the end of the pull URL:

https://patch-diff.githubusercontent.com/raw/linux-audit/audit-userspace/pull/23.patch

stevegrubb:
I rewrote part of the patch to use strdup to store paths. I also did not pull in the input command line option. The main reason being is that during the 2.9 development cycly audispd is going to be absorbed into auditd and it won't make sense to add something and then take it away. :-)  Thanks for the patch. I gave you credit in the Changelog.

PR #25, opened 2017-06-20
Fix build with musl libc

doughdemon:
Some patches fixing compilation with musl libc.There also seems to be some bad interaction between musl's <stdint.h> and SWIG (see https://bugs.gentoo.org/620006#c4). Would it be possible to remove the <stdint.h> %include in bindings/swig/src/auditswig.i or to change it into an %import?

### DIFF
--- a/audisp/audispd.c
+++ b/audisp/audispd.c
@@ -34,6 +34,7 @@
 #include <sys/poll.h>
 #include <netdb.h>
 #include <arpa/inet.h>
+#include <limits.h>
 
 #include "audispd-config.h"
 #include "audispd-pconfig.h"
--- a/auparse/auparse.c
+++ b/auparse/auparse.c
@@ -1102,10 +1102,12 @@ static int extract_timestamp(const char *b, au_event_t *e)
 	int rc = 1;
 
         e->host = NULL;
+
+	tmp = alloca(340);
 	if (*b == 'n')
-		tmp = strndupa(b, 340);
+		tmp = strncpy(tmp, b, 340);
 	else
-		tmp = strndupa(b, 80);
+		tmp = strncpy(tmp, b, 80);
 	ptr = audit_strsplit(tmp);
 	if (ptr) {
 		// Optionally grab the node - may or may not be included
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -819,10 +819,9 @@ static const char *print_proctitle(const char *val)
 		// Proctitle has arguments separated by NUL bytes
 		// We need to write over the NUL bytes with a space
 		// so that we can see the arguments
-		while ((ptr  = rawmemchr(ptr, '\0'))) {
-			if (ptr >= end)
-				break;
-			*ptr = ' ';
+		while (ptr < end) {
+			if (*ptr == '\0')
+				*ptr = ' ';
 			ptr++;
 		}
 	}
--- a/src/auditctl.c
+++ b/src/auditctl.c
@@ -32,6 +32,8 @@
 #include <ctype.h>
 #include <unistd.h>
 #include <sys/utsname.h>
+#include <sys/select.h>
+#include <sys/time.h>
 #include <fcntl.h>
 #include <errno.h>
 #include <libgen.h>	/* For basename */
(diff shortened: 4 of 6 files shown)


stevegrubb:
I applied the 1st and 4th patch. Thanks. The other two I'll leave for the time being.

tych0:
Thanks! Is there something I can do to make the other two patches acceptable?

stevegrubb:
Both strdupa and rawmemchr are used for performance. I might revisit this at some point when I do more benchmarking and optimization. This would possibly be in the 2.9 series of releases.

tych0:
Ok, thanks. Let me know if I can help.

stevegrubb:
I merged the 5th chunk of this patch. I have a question, does musl libc have alloca but not strdupa?

stevegrubb:
I just looked at musl libc source and it seems to have strdupa as a define:

http://git.musl-libc.org/cgit/musl/tree/include/string.h#n89

Does the audit package build OK these days? I'm wondering if this is new and older copies of mucl libc have a problem.

stevegrubb:
Closing this out. I've taken in as much of the patch as I think is feasible. If anyone disagrees, reopen this with a new patch. Thanks for the suggestions!

bluca:
@stevegrubb hi, just tested with the very last muslc, 1.1.21, and this still fails:

```
libtool: link: /tmp/musl/bin/musl-gcc -fPIE -DPIE -g -D_REENTRANT -D_GNU_SOURCE -fno-strict-aliasing -pthread -Wno-pointer-sign -g -O2 -pie -Wl,-z -Wl,relro -Wl,-z -Wl,now -o .libs/auditd auditd-auditd.o auditd-auditd-event.o auditd-auditd-config.o auditd-auditd-reconfig.o auditd-auditd-sendmail.o auditd-auditd-dispatch.o auditd-auditd-listen.o  -Llibev -lev -L../audisp -ldisp -L../lib /tmp/audit-userspace/lib/.libs/libaudit.so -L../auparse /tmp/audit-userspace/auparse/.libs/libauparse.so -lpthread -lrt -lm -L../common -laucommon -pthread -Wl,-rpath -Wl,/tmp/musl/lib
/tmp/audit-userspace/auparse/.libs/libauparse.so: undefined reference to `rawmemchr'
/tmp/audit-userspace/auparse/.libs/libauparse.so: undefined reference to `strndupa'
```

As you pointed out musl does have strdupa, but the code here is using str**n**dupa which is missing. Same for rawmemchr.
The fixes from this PR work - any chance they could be picked up for the next release? Or is there a different solution you would like to see implemented?

Thank you!

bluca:
To quickly reproduce:

```
cd /tmp
wget https://www.musl-libc.org/releases/musl-1.1.21.tar.gz
tar xf musl-1.1.21.tar.gz
cd musl-1.1.21
./configure --prefix=/tmp/musl --syslibdir=/tmp/musl/lib
make -j5 install
ln -sr /usr/include/asm /tmp/musl/include/
ln -sr /usr/include/asm-generic/ /tmp/musl/include/
ln -sr /usr/include/linux /tmp/musl/include/
cd ../audit-userspace
./autogen.sh
/configure --prefix=/tmp/musl LD_LIBRARY_PATH=/tmp/musl/lib CC=/tmp/musl/bin/musl-gcc --disable-zos-remote
make -j5
```

stevegrubb:
The rawmemchr can be replaced with a call to strchr. And strndupa can be replaced with a macro that does alloca and strncpy. Maybe that is the right approach? Then I can put something in configure.ac that detects if those two functions exist and swing in the right defintion.

bluca:
Sounds great! Happy to help with testing if needed. Thanks!

stevegrubb:
Commit d579a08 should have a fix for this.

bluca:
@stevegrubb thank you, works like a charm!

CameronNemo:
I had issues with trying to build the zos-remote plugin on musl. `pthread_yield()` was not defined.

doughdemon:
`pthread_yield()` is non-standard, the plugin should use `sched_yield()` instead.

stevegrubb:
Commit a5e404a should fix this. If there are other portability problems, please open a new issue. Thanks!

PR #27, opened 2017-07-25
fix auparse serial parsing of event when system time < 9 characters

kruvin:
If the system time is low, semicolon indicating serial is not found in string due to +9 optimization, e.g. in:

type=CWD msg=audit(215.159:13046):  cwd="/"

### DIFF
--- a/auparse/auparse.c
+++ b/auparse/auparse.c
@@ -1073,14 +1073,7 @@ static int str2event(char *s, au_event_t *e)
 	char *ptr;
 
 	errno = 0;
-	ptr = strchr(s+10, ':');
-	if (ptr) {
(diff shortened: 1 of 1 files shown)


stevegrubb:
Is this for records where the system clock has gotten reset to 1970?

kruvin:
Yes, for example on some embedded devices that only set their time a few seconds after booting

kruvin:
Notice that the optimization is meaningless, since there's an "strchr" for the dot ('.') right after. It can be optimized to take the position of the semicolon (':') at +4 after the dot.

stevegrubb:
Applied. Thanks!

PR #36, opened 2017-12-30
expose all members of ausparse_type_t to Python

benjaminp:
Some new record types have been added recently, but the Python bindings haven't kept up.


stevegrubb:
Ah! Good catch. Thanks!

PR #38, opened 2018-02-21
Do not rotate logs when num_logs < 2.

brozs:
Previously, the check was for (max_log_size_action == SZ_ROTATE && config->num_logs < 2) only. However, the rotate_logs() function can be invoked using a USR1 signal independently on the action defined.

When num_logs = 0 and max_log_file_action = IGNORE (other than ROTATE) the auditd busy-looped due to integer underrun in the for loop.
With num_logs = 1 and max_log_file_action = IGNORE, the oldname was uninitialized.

This commit exits the rotate_logs() function when log rotation is disabled and adds a LOG_NOTICE.

### DIFF
--- a/src/auditd-event.c
+++ b/src/auditd-event.c
@@ -968,9 +968,14 @@ static void rotate_logs(unsigned int num_logs)
 	unsigned int len, i;
 	char *oldname, *newname;
 
-	if (config->max_log_size_action == SZ_ROTATE &&
-				config->num_logs < 2)
+	/* Check that log rotation is enabled in the configuration file. There is
+	 * no need to check for max_log_size_action == SZ_ROTATE because this could be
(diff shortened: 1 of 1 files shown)


stevegrubb:
Just one quick question. What are the steps to reproduce the problem this is fixing?

brozs:
Reproduction steps:
1. Edit /etc/audit/auditd.conf so that:
    num_logs = 0
    max_log_file_action = ignore
2. Reload configuration:
    # systemctl reload auditd
3. Invoke log rotation:
    # kill -USR1 $(pidof auditd)

Result: auditd in a busy loop.

I am unsure about the LOG_NOTICE - isn't there any situation where that would become flooding the logs annoyingly?

brozs:
I am also thinking about reopening the log file even when num_logs < 2. External log rotation could be performed that way easily.

PR #39, opened 2018-02-22
Convert auparse_test to run with python3

scarabeusiv:
This allows only python3 usage for both build and runtime

### DIFF
--- a/auparse/test/auparse_test.py
+++ b/auparse/test/auparse_test.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 import os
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for this patch! I am going to park it for a little while and come back to it. I'm currently in the 2.8 series which is aimed at systems supporting python2. Somewhere around 2.9/3.0 I plan to make changes that are not backwards compatible and that would be the right time for a change like this. The patch looks good other wise.

scarabeusiv:
It is python2 compatible, there is no problem in py2 only execution (2.7+)

stevegrubb:
Thanks for the patch!

PR #44, opened 2018-04-10
ausearch: Defend against parsing bugs

WOnder93:
Prepend all search string of type "\<name\>=" with a space to prevent
bugs arising from different attributes with matching suffixes (e.g.
'uid' and 'auid').

Even the current code contains at least one such bug -- without this
patch, the following audit event is matched with no options, but not
matched with '-ui 0':

```
type=SYSCALL msg=audit(1523350632.356:24): arch=c000003e syscall=87 success=yes exit=0 a0=55d3d55ed2e0 a1=7ffec3e73380 a2=7ffec3e73380 a3=0 items=2 ppid=1452 pid=1455 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=ttyS0 ses=1 comm="perl" exe="/usr/bin/perl" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)
```

There is still some concern with attributes like 'new something=...'
as conflict with these won't be caught even with this patch.

### DIFF
--- a/src/ausearch-parse.c
+++ b/src/ausearch-parse.c
@@ -222,7 +222,7 @@ static int parse_task_info(lnode *n, search_items *s)
 
 	// ppid
 	if (event_ppid != -1) {
-		str = strstr(term, "ppid=");
+		str = strstr(term, " ppid=");
 		if (str != NULL) { // ppid is an optional field
 			ptr = str + 5;
 			term = strchr(ptr, ' ');
@@ -254,9 +254,9 @@ static int parse_task_info(lnode *n, search_items *s)
 	}
 	// optionally get loginuid
 	if (event_loginuid != -2 || event_tauid) {
-		str = strstr(term, "auid=");
+		str = strstr(term, " auid=");
 		if (str == NULL) {
-			str = strstr(term, "loginuid=");
+			str = strstr(term, " loginuid=");
 			if (str == NULL)
 				return 19;
 			ptr = str + 9;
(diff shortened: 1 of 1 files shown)


WOnder93:
To reproduce the bug mentioned in the commit message:
```bash
#!/bin/bash

RECORD='type=SYSCALL msg=audit(1523350632.356:24): arch=c000003e syscall=87 success=yes exit=0 a0=55d3d55ed2e0 a1=7ffec3e73380 a2=7ffec3e73380 a3=0 items=2 ppid=1452 pid=1455 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=ttyS0 ses=1 comm="perl" exe="/usr/bin/perl" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)'

echo "Without -ui 0:"
ausearch -if <(echo $RECORD) || exit 1

echo "With -ui 0:"
ausearch -if <(echo $RECORD) -ui 0 || exit 1
```
I get the following output:
```
Without -ui 0:
----
time->Tue Apr 10 10:57:12 2018
type=SYSCALL msg=audit(1523350632.356:24): arch=c000003e syscall=87 success=yes exit=0 a0=55d3d55ed2e0 a1=7ffec3e73380 a2=7ffec3e73380 a3=0 items=2 ppid=1452 pid=1455 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=ttyS0 ses=1 comm="perl" exe="/usr/bin/perl" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)
With -ui 0:
<no matches>
```

stevegrubb:
After looking over the patch, I think I'd prefer a more conservative approach and just fix the issue pointed out by the reproducer. By lengthening each match, it will impact performance and there will be a lot of matches on the space being added - which will cause additional slowdown. Thanks.

PR #45, opened 2018-04-10
Add support for executable name not equal operator

WOnder93:
Related kernel GH issue: https://github.com/linux-audit/audit-kernel/issues/53

### DIFF
--- a/lib/libaudit.c
+++ b/lib/libaudit.c
@@ -1585,8 +1585,8 @@ int audit_rule_fieldpair_data(struct audit_rule_data **rulep, const char *pair,
 				uint32_t features = audit_get_features();
 				if ((features & AUDIT_FEATURE_BITMAP_EXECUTABLE_PATH) == 0)
 					return -EAU_FIELDNOSUPPORT;
-				if (op != AUDIT_EQUAL)
-					return -EAU_OPEQ;
+				if (!(op == AUDIT_NOT_EQUAL || op == AUDIT_EQUAL))
+					return -EAU_OPEQNOTEQ;
 				_audit_exeadded = 1;
 			}
 			if (field == AUDIT_FILTERKEY &&


pcmoore:
@stevegrubb before I merge the kernel patch (see PR above, includes mailing list pointers) I want to make sure you are going to merge the corresponding userspace patch.  Any updates at this point?

pcmoore:
@stevegrubb thanks for looking at the patch.  To answer the "why" question, @rgbriggs was the one who originally captured this request as a GH issue (linked above: linux-audit/audit-kernel#53) based on a mailing list posting (in the kernel issue: [archive link](https://www.redhat.com/archives/linux-audit/2017-June/msg00029.html)).

stevegrubb:
@pcmoore Thanks. OK, I see where this come from. The problem that the user was mentioning is the well known problem where a daemon on RHEL 6 and lower pickup the auid of the user who starts the daemon. Moving to a systemd based OS is the only fix for this problem that I know of. 

So, I guess the use case for this is when you have one specific program that is supposed to have exclusive access a directory and you want to catch anyone else that may try to access it. But it leaves unsolved the case where you have several processes that you want to filter out. For example, you may have a writer process and a reader process that should have exclusive access...no solution for this.

WOnder93:
Hm... now that I think about it, wouldn't all the problems be resolved by allowing the exclude filter to filter by executable name? In fact, there already seems to be a kernel issue filed about this (https://github.com/linux-audit/audit-kernel/issues/54). Or am I misunderstanding how the exclude filter works?

rgbriggs:
On 2018-04-20 08:27, Ondrej Mosnáček wrote:
> Hm... now that I think about it, wouldn't all the problems be resolved by allowing the exclude filter to filter by executable name? In fact, there already seems to be a kernel issue filed about this (https://github.com/linux-audit/audit-kernel/issues/54). Or am I misunderstanding how the exclude filter works?

Yes, this was part of the motivation for filing that issue.

WOnder93:
OK, so @stevegrubb, @pcmoore - would you prefer that I drop the `exe!=` patches and switch to implementing https://github.com/linux-audit/audit-kernel/issues/54 instead?

rgbriggs:
On 2018-04-20 11:08, Ondrej Mosnáček wrote:
> OK, so @stevegrubb, @pcmoore - would you prefer that I drop the
> `exe!=` patches and switch to implementing
> https://github.com/linux-audit/audit-kernel/issues/54 instead?

Is there any reason not to do both?  Having this option not available
would seem surprising to me if I were trying to write rules.

pcmoore:
I agree with @rgbriggs, there doesn't seem to be any harm in implementing both, while only implementing one could leave someone unhappy.

FYI, @WOnder93 I went ahead and assigned linux-audit/audit-kernel/issues/54 to you as well.

PR #46, opened 2018-04-17
ausearch: Fix incorrect matching of 'uid' field

WOnder93:
When auid is not searched for, the parser matches the 'auid' field as
'uid' and skips the actual 'uid' field. This patch fixes it by
searching for " uid=" instead of "uid=".

This is triggered for example on the following record:

```
type=SYSCALL msg=audit(1523350632.356:24): arch=c000003e syscall=87 success=yes exit=0 a0=55d3d55ed2e0 a1=7ffec3e73380 a2=7ffec3e73380 a3=0 items=2 ppid=1452 pid=1455 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=ttyS0 ses=1 comm="perl" exe="/usr/bin/perl" subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null)
```

### DIFF
--- a/src/ausearch-parse.c
+++ b/src/ausearch-parse.c
@@ -276,7 +276,7 @@ static int parse_task_info(lnode *n, search_items *s)
 	}
 	// optionally get uid
 	if (event_uid != -1 || event_tuid) {
-		str = strstr(term, "uid=");
+		str = strstr(term, " uid=");
 		if (str == NULL)
 			return 22;
 		ptr = str + 4;
@@ -824,7 +824,7 @@ static int parse_user(const lnode *n, search_items *s)
 	}
 	// optionally get uid
 	if (event_uid != -1 || event_tuid) {
-		str = strstr(term, "uid=");
+		str = strstr(term, " uid=");
 		if (str == NULL)
 			return 4;
 		ptr = str + 4;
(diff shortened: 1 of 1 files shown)


stevegrubb:
This patch also looks bigger than the reproducer. Also, even when I add just the " uid" part for the task_info_parser, it fails the ausearch-test suite. I don't know if you have a good sized collection of events, but if you do, try checking the patch against the test suite. The parsers are sensitive to changes.

stevegrubb:
I fixed this a different way in patch  ec35efa.  Thanks for pointing out the issue.

rgbriggs:
On 2018-04-19 14:23, Ondrej Mosnáček wrote:
> I tried to look for the "ausearch-test suite", but I couldn't find anything like that in this repository... there are some au*parse* tests and two unit tests, but nothing that would directly test ausearch. Did you mean the [audit-testsuite](https://github.com/linux-audit/audit-testsuite/)? (I think it is intended for the kernel, but it indirectly tests ausearch as well.) If not, could you point me where I could find it?

I think you are (reasonably) looking for this:
	https://github.com/linux-audit/audit-documentation/issues/10

For now it is still in:
	https://people.redhat.com/sgrubb/audit/
	https://people.redhat.com/sgrubb/audit/ausearch-test-0.6.tar.gz

PR #48, opened 2018-04-25
Allow exclude and user filter by executable name

WOnder93:
This patch removes the restriction that excludes the AUDIT_EXE field
from the exclude filter list. It also clarifies in the documentation
that this field can be used also with these filter lists ('auditctl -a
user,always -F exe=/something' works even without this patch).

Relevant kernel patch that enables exclude filter for AUDIT_EXE:
https://www.redhat.com/archives/linux-audit/2018-April/msg00114.html

Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>


pcmoore:
@stevegrubb can we get a thumbs up, or down, on the userspace piece?  I don't want to merge the kernel piece only to find out you've got objections to the userspace portion.

pcmoore:
@stevegrubb can we get a verdict on the userspace part of this?  The kernel is at -rc4 right now, and if we want this to go in during the next merge window this should be merged into audit/next within the next week or so.

pcmoore:
Pinged @stevegrubb directly via email in case he isn't monitoring his GitHub notifications.

PR #61, opened 2018-08-24
Fix memory leak in audisp-remote

truthsolo:
Looks like etok.value should be free'd here


truthsolo:
No problem, glad I could help. Filed it here for tracking https://bugzilla.redhat.com/show_bug.cgi?id=1622194

Issue #58, opened 2018-08-01
auditd and message format: unescaping


dimv36:
Hi, I asked for escaping non-standart ASCII raws in #56 previously.

Now I wrote messages into auditd daemon and want to decode them out.
Auditd' Python API wrote me unescaped values like this:

event=D09FD0BED0B4D0BAD0BBD18ED187D0B5D0BDD0B8D0B5

how can I decode them?

rgbriggs:
On 2018-08-01 14:32, Dmitry Voronin wrote:
> Hi, I asked for escaping non-standart ASCII raws in #56 previously.
> 
> Now I wrote messages into auditd daemon and want to decode them out.
> Auditd' Python API wrote me unescaped values like this:
> 
> event=D09FD0BED0B4D0BAD0BBD18ED187D0B5D0BDD0B8D0B5
> 
> how can I decode them?

ausearch -i

dimv36:
Unfortunately, it's not working for me. Does libaudit work fine for non-ascii symbols?

dimv36:
Guys,

I try to encode my non-ASCII value (and ASCII value with spaces) via audit_encode_nv_string() and send to audit log successfully. I used an event type AUDIT_TRUSTED_APP.

But I can't get decoded value via ausearch -i or Python audit API. Please, help!
I'm using libaudit 2.8.3.

dimv36:
I found, that attrs, which are needed to be unescaped, are determined in auparse_interp_adjust_type().

So, I make a simple patch for me for unescaping those attrs for my case depended on event_type AUDIT_TRUSTED_APP. But I don't think, that good idea. As for me, configurable list of attributes which must be unescaped is best solution for my problem.

So, my proposal:
* Add configuration value in aduitd.conf (?) for list of unescaped attributes in event;
* Load those values into list and check attribute name in auparse_interp_adjust_type() for unescaping
* If attribute in list, use au_unescape

stevegrubb:
Yes, that sounds like a good idea. But I am not sure about where to put the field list. Ausearch needs the information and nothing else does. So, it might be a command line option like --escape-list. Then you can make a bash alias so that anytime you run ausearch it automatically uses the list. It would also need to apply only to AUDIT_TRUSTED_APP and AUDIT_KERNEL_OTHER records to ensure it cannot be abused for other records.

dimv36:
I'm not think, that only ausearch is needed in those list.
[...]

PR #67, opened 2018-09-05
Support for unencoding values from TRUSTED_APP

dimv36:
How I notice in https://github.com/linux-audit/audit-userspace/issues/58

I needed in sending custom messages (with spaces, not ASCII attribute values) in audit log.
So, I made it, but next I needed for unescaping those values.

I can't do this via current API, so I made a simple patch, that allows me determine if value needed in unescaping. If needded auparse use au_unescape() function to do this.

I abandoned the original idea, because my changes are more simple to solve my problem.
I think, that auparse needed in more intellectual API for determine field's type.

### DIFF
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -312,6 +312,16 @@ static void key_escape(char *orig, char *dest, auparse_esc_t escape_mode)
 	}
 }
 
+static int is_int_string(const char *str)
+{
+	while (*str) {
+		if (!isdigit(*str))
+			return 0;
+		str++;
+	}
+	return 1;
+}
+
 static int is_hex_string(const char *str)
 {
 	while (*str) {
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for the patch. I will give feedback or merge it next week.

auparse/interpret.c
```diff
@@ -2876,6 +2886,20 @@ int auparse_interp_adjust_type(int rtype, const char *name, const char *val)
 		(rtype == AUDIT_ADD_GROUP || rtype == AUDIT_GRP_MGMT ||
 			rtype == AUDIT_DEL_GROUP))
 		type = AUPARSE_TYPE_GID;
+	else if (rtype == AUDIT_TRUSTED_APP) {
+		if (val[0] == '"')
+			type = AUPARSE_TYPE_ESCAPED;
+		else if (strcmp(name, "id") == 0)
```
stevegrubb: Why are we singling out "id" as needing interpretation? If we do that one, then we need to do more. Perhaps just detecting escaped for trusted app records is all that should be done. It should somehow call lookup_type otherwise.

auparse/interpret.c
```diff
@@ -2876,6 +2886,20 @@ int auparse_interp_adjust_type(int rtype, const char *name, const char *val)
...
+			type = AUPARSE_TYPE_UID;
+		else if (is_int_string(val))
+			type = AUPARSE_TYPE_UNCLASSIFIED;
+		/* Check if we have string with only HEX symbols,
+		 * but except HEX strings like 0xABCD */
+		else if (is_hex_string(val) && (strncmp(val, "0x", 2) != 0))
```
stevegrubb: Starting a value with 0x has never been part of the audit coding standard for escaped fields. I don't think we should expand it to this format.

stevegrubb:
OK. I looked over the patch. I have some comments inline.

auparse/interpret.c
dimv36: OK, agreed.

auparse/interpret.c
dimv36: This is prevents possible unescaping for integer values for id-like fields.

stevegrubb: Right. But what I'm thinking is that it should be something like 
```
else if (rtype == AUDIT_TRUSTED_APP) {
        if (val[0] == '"')
                type = AUPARSE_TYPE_ESCAPED;
        else
                type = lookup_type(name);
}
```
This way it looks up more than just "id".

dimv36: No, thanks.

stevegrubb:
Going to close this. Thanks for your interest in improving the audit system. If the above suggested changes were made, I'd consider merging this. Otherwise, we should close it since it's not ready.

PR #69, opened 2018-10-01
bindings/swig/src/auditswig.i: Do not hardcode the path of stdint.h

bigon:
auditswig.i hard codes the path to stdint.h. That will fail to work with
non-glibc libcs and after moving glibc's headers (#798955). The path is
hard coded, because swig's %include does not search the standard header
search path. Rather than using %include here, we can use #include,
because stdint.h does not declare any functions. Thus swig entirely
ignores stdint.h and leaves the search to the C compiler.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=909967


stevegrubb:
One question, how do you feel about the hardcoded audit.h right above this one?

bigon:
I think it's less a problem because the header is the same across environment

PR #79, opened 2019-01-11
auditd.service: use /run/auditd.pid instead of /var/run/auditd.pid

Maryse47:
`/var/run/` is a legacy path which is currently used as symlink to `/run`. Newer systemd will update this path themselves and warn about it:

> /usr/lib/systemd/system/auditd.service:12: PIDFile= references path below legacy directory /var/run/, updating /var/run/auditd.pid → /run/auditd.pid; please update the unit file accordingly.

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -17,7 +17,7 @@ Documentation=man:auditd(8) https://github.com/linux-audit/audit-documentation
 
 [Service]
 Type=forking
-PIDFile=/var/run/auditd.pid
+PIDFile=/run/auditd.pid
 ExecStart=/sbin/auditd
 ## To not use augenrules, copy this file to /etc/systemd/system/auditd.service
 ## and comment/delete the next line and uncomment the auditctl line.


stevegrubb:
First, I'd like to thank you for your interest in make the audit project better. The audit daemon support many varieties of linux distributions. I run audit-3.0 on an old RHEL6 instance. It works just fine on those old systems. Those old systems do not have /run. Audit still supports sysvinit which is what you get when you don't pass --enable-systemd to ./configure. Maybe in a few years we can drop that support. But I'd think we should move to /run/ at the same time sysvinit is also deprecated. Again, thanks for the suggestion. If you have any other ideas on improving audit, please share.

Maryse47:
Ok, that sounds reasonable. I hope you release 3.0 soon so everyone could use it 😃 .

stevegrubb:
3.0 is waiting on the kernel container work to land. It was the major feature that was to distinguish 3.0 from the 2.0 series. I'm hoping the container work is accepted by March. In the meantime, I have pushed out at least one pre-release.

PR #81, opened 2019-01-17
Update all paths /var/run -> /run

eworm-de:
In recent versions systemd starts complaining:

/usr/lib/systemd/system/auditd.service:12: PIDFile= references path below
legacy directory /var/run/, updating /var/run/auditd.pid → /run/auditd.pid;
please update the unit file accordingly.

So let's just update the paths.

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -17,7 +17,7 @@ Documentation=man:auditd(8) https://github.com/linux-audit/audit-documentation
 
 [Service]
 Type=forking
-PIDFile=/var/run/auditd.pid
+PIDFile=/run/auditd.pid
 ExecStart=/sbin/auditd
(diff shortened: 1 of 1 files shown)


stevegrubb:
First, I'd like to thank you for your interest in make the audit project better. The audit daemon supports many varieties of linux distributions. I run audit-3.0 on an old RHEL6 instance. It works just fine on those old systems. Those old systems do not have /run. Audit still supports sysvinit which is what you get when you don't pass --enable-systemd to ./configure. Maybe in a few years we can drop that support. But I'd think we should move to /run/ at the same time sysvinit is also deprecated. I would accept updating the systemd service & state files since we know they are just /run/. Again, thanks for the suggestion. If you have any other ideas on improving audit, please share.

eworm-de:
Updated my commit to only change the path in `init.d/auditd.service`.

PR #83, opened 2019-01-30
fix example to not reply on strnlen to get data size

cheese:


stevegrubb:
Hello, thanks for helping to make the audit project a little more robust. Looking at the patch, do you mind explaining a little more about why this patch is needed?

cheese:
I did not dive too deep. I just followed the original example to create a plugin for a simple need. But the callback was not called at once when message had come. The issue was gone after applying this patch. I am on CentOS 7.

With direct respect to the code, there are two issues of the original code that this patch intends to fix:
1. Using strlen to get the length of buffer read by fgets may be broken if there is null byte within the buffer. The return code of read can be simply the buffer length. No extra strlen call is needed.
2. Using fgets requires an extra byte for the ending null byte. Not needed for direct 'read' call.

stevegrubb:
I'm going to apply this fix and then make some additional changes. I don't think we want it to be non-blocking as that causes a whole other error (EAGAIN) that needs to be handled and busy looping is not ideal. I think we want to add select to the mix. Thanks for the patch!

stevegrubb:
Actually, it looks good as is.

PR #84, opened 2019-03-14
Clarify the documentation for the space_left parameter.

qralston:
Specifically, clarify that the audit daemon triggers ```space_left_action``` when the free space in the filesystem containing ```log_file``` drops below the value specified by ```space_left``` (either as an absolute value, or as a percentage).

The previous description did not make it clear exactly how the ```space_left``` parameter was interpreted, leading to confusion.


stevegrubb:
This is a nice addition. Thanks.

PR #88, opened 2019-03-22
Harden systemd service

topimiettinen:
Signed-off-by: Topi Miettinen <toiwoton@gmail.com>

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -33,6 +33,26 @@ MemoryDenyWriteExecute=true
 LockPersonality=true
 ProtectControlGroups=true
 ProtectKernelModules=true
+CapabilityBoundingSet=CAP_AUDIT_CONTROL CAP_SYS_NICE CAP_SYS_ADMIN
+DeviceAllow=/dev/null rw
+DevicePolicy=strict
+IPAddressDeny=any
+LimitMEMLOCK=0
+PrivateDevices=yes
+PrivateNetwork=yes
+PrivateTmp=yes
+ProtectHome=yes
+ProtectKernelTunables=yes
+ProtectSystem=strict
+ReadOnlyPaths=/
+ReadWritePaths=/proc/self /run /tmp /var/log/audit
+RestrictAddressFamilies=AF_UNIX AF_NETLINK
+RestrictNamespaces=yes
+RestrictRealtime=yes
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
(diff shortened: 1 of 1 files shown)


stevegrubb:
Hello,

First, thanks for looking at making auditd safer/better. The audit daemon can do remote logging via the audisp-remote plugin. It can also aggregate logs sent to it by other audit daemons. Does this systemd policy get inherited by its children? Would this cause third party plugins to have problems doing whatever then need to do?

topimiettinen:
Yes, the children are also restricted. Unfortunately I didn't consider that case, so I'll have to make this much less restrictive. Are there any limits on what the plugins could do? For example, could a plugin reasonably turn off the system in the name of security (thus requiring `@reboot` class of system calls and related `CAP_SYS_BOOT` capabilities)?

stevegrubb:
That is one of the problems, the plugins could do anything. I do not have a full list of all the plugins open source or commercial. And I have some plugin in development that do things like manipulate iptables or kill processes/sessions. If you could divide this into things that should be safe for any use case and things that might be more speculative, then we can look at each on it's own merit. I'm open to improvements, but plugins might do all kinds of things.

topimiettinen:
I don't know SELinux well, but it seems that its "logging" module allows audispd to launch shell, MTA and other executables. More capabilities are also available. That would indicate that it's hard to constrain the plugins much, if at all.

I think the mildest directive in the list is `SystemCallArchitectures=native`, which restricts the system call interface to native versions only (e.g. x86-64 on x86-64, but not i386 or x32). Could it be reasonable to expect use of 32 bit plugins on a 64 bit system?

stevegrubb:
I would not expect a 32 bit plugin on a 64 bit system. There's a small chance it might occur in the commercial space. But even then, I'd still expect a native 64 bit application.

topimiettinen:
Actually, even `SystemCallArchitectures=native` (like many other nice settings) automatically sets `NoNewPrivileges=yes`, which forbids executing any setuid/setgid binaries. I know that at least the notorious `sendmail` was setuid, so even this setting might cause problems.

I'm not sure if there's anything to be done to harden in the general case. Perhaps there would be more opportunities if the architecture could be changed so that audispd could be configured to be launched as a separate systemd service instead of child of auditd?

stevegrubb:
OK. Thanks for looking into this. I guess there is nothing that can be done here. If you have any other improvements, please open an issue or pull request.

PR #92, opened 2019-04-17
Support for unescaping in trusted messages. Take 2

dimv36:
Hi, how I write in https://github.com/linux-audit/audit-userspace/pull/67, this is adds support for unescaping messages via au_unescape()

### DIFF
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -454,6 +454,7 @@ int load_interpretation_list(const char *buffer)
 				*ptr = 0;
 			} else
 				tmp = 0;
+
 			n.val = strdup(val);
 			nvlist_append(&il, &n);
 			nvlist_interp_fixup(&il);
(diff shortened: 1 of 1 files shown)


stevegrubb:
I took a quick look. I still don't like the strcmp on the id field. That will make it not generic. But otherwise I think it looks OK. There is a new function added is_int_string(). Is this because some values are decimal rather than hex and it was being picked up as hex?

dimv36:
is_int_string() is used for determining if field value contains only hex symbols, yes.

As for using strcmp(), I can use something else for this cases.

stevegrubb:
I am going to commit this a different way instead of merging a pull request. I like the patch except the strcmp on id. I will remove that part and commit it.

stevegrubb:
This went in as commit 10c64d5. Thanks!

PR #95, opened 2019-04-25
Port auditd.init and augenrules to the POSIX shell

bigon:


dirtyharrycallahan:
prefer signal names instead of numbers for readability
prefer $(...) instead of legacy backticks

PR #97, opened 2019-05-03
auditd.service (systemd startup file): enable reload

otheus:
This enables systemctl to reload the service in the expected manner: auditd reloads its configuration, and the rules are reloaded via augenrules. This is expected behavior and required for SysCMs such as puppet which link a service to a file-change.

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -28,6 +28,9 @@ ExecStartPost=-/sbin/augenrules --load
 # the next line after copying the file to /etc/systemd/system/auditd.service
 #ExecStopPost=/sbin/auditctl -R /etc/audit/audit-stop.rules
 
+ExecReload=-/sbin/augenrules --load
+ExecReload=/bin/kill -HUP $MAINPID
+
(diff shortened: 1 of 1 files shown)


stevegrubb:
Hi. First, Thanks for submitting an improvement for the audit system.

I am wondering if it would be better to make this modification in init.d/auditd.reload. The problem for auditd is that we have to have an audit trail of any user that tries to alter the audit trail. Anything coming from systemd itself has an auid of -1 which means we have no idea who submitted the request. For this reason, we MUST run all commands through "service" which issues the request in the user's context so that the auid is correctly set. This is a basic security requirement across many security standards such as Common Criteria, PCI-DSS, NISPOM, etc that we must follow.

I would be willing to merge a patch against init.d/auditd.reload if you want to create one.

stevegrubb:
Really wished you had made the change so that you got credit for the suggestion. I went ahead and fixed this in the legacy reload script as commit 9e830ed. Thanks for pointing out the issue.

PR #102, opened 2019-08-02
Autotools: Use Python 2 explicitly

hroncok:
Bu copypasting the Python 3 parts and bending it, until it builds.

This has only seen very limited testing.


besser82:
@hroncok Do you agree with me #103 is less intrusive and likely less error prone?

hroncok:
Yes.

PR #104, opened 2019-08-08
Add support to monitor BPF program loads/unloads

olsajiri:
hi,
we would like to add BPF program tracking into audit daemon,
so we could monitor  BPF programs loading and unloading
in the systerm.

This pull request is **RFC** to get an idea from you guys how you
feel about this and what perhaps needs to be made differently
or is missing.

thanks for the feedback,
jirka

---
Adding src/auditd-bpf.c object with BPF related code that:
  - creates perf event for every cpu in the system
  - creates memory map for each event
  - adds perf events file descriptor into the global event
    loop object with handler

The handler function reads the events from memory map 
and calls send_audit_event on BPF events generating
following audit messages on BPF program load/unload:

```
BPF audit initialized, rb size 655360
Init complete, auditd 3.0 listening for events (startup state enable)
...
type=BPF msg=audit(1565274022.716:2986): pid=13739 tid=13739 comm=execsnoop perf-time=55136.376792231 prog-id=119 event=LOAD
type=BPF msg=audit(1565274022.725:2987): pid=13739 tid=13739 comm=execsnoop perf-time=55136.385981813 prog-id=120 event=LOAD
...
type=BPF msg=audit(1565274030.707:2988): pid=13739 tid=13739 comm=N/A perf-time=55144.367828929 prog-id=120 event=UNLOAD
type=BPF msg=audit(1565274030.707:2989): pid=13739 tid=13739 comm=N/A perf-time=55144.367840418 prog-id=119 event=UNLOAD

```
Because of events processing delay we might not be able
to see process name (read from /proc/<PID>/comm), so the 
comm might have the 'N/A' value sometimes.

I added `AUDIT_BPF` type, but I'm not too sure about the 
proper number for it, perhaps this should be discussed.

The memory maps should be pretty low traffic and the 
default size (4 pages per event) should be enough for 
most setups.

However there's a possibility to setup perf memory map 
size with new auditd's config file option:
`  bpf_mem_pages = <NUM>`

It's the number of pages that will be used as size for 
perf event memory map size and must be power of 2. It's
possible to add more user friendly option for setting
the memory size.

It's possible to disable BPF code with configure option:
`  --disable-bpf`


stevegrubb:
The new patch looks fine. Will merge when Kernel piece is accepted.

olsajiri:
> The new patch looks fine. Will merge when Kernel piece is accepted.

should I make PR for it? plz let me know

stevegrubb:
Looks like you did a force push and that replaced the previous patch. Meaning, no changes needed.

PR #107, opened 2019-09-27
Further simple enhancement to contrib/plugin/audisp-example.c

cheese:


stevegrubb:
I can see merging the  "audisp-example: remove the left 'strnlen', completes e9500ed" patch. I'd merge that on its own. Thanks for pointing that out.

I don't really want to keep extending the example program to do something specific. If it were demonstrating a libaudit or libauparse feature, that might be appropriate to add. But writing to a file is easy enough others can figure that out without needing an example.

stevegrubb:
If you want to resubmit just the e9500ed commit as a pull request, I'd be happy to merge it.

PR #115, opened 2020-02-07
Add support for mips 32-bit & 64-bit (o32,n32,n64)

RayWang10:
testing was done on loongson3a(mips little-endian) using debian,
    kernel 4.19 with patch
    https://www.linux-mips.org/archives/linux-mips/2018-12/msg00193.html


lib/libaudit.h
```diff
@@ -36,6 +36,58 @@ extern "C" {
...
+ * the values.
+ */
+#ifndef AUDIT_ARCH_MIPS64EL
+#define AUDIT_ARCH_MIPS64EL    (EM_MIPS|__AUDIT_ARCH_64BIT|__AUDIT_ARCH_LE)
+#endif
+#ifndef __AUDIT_ARCH_ALT
+#define __AUDIT_ARCH_ALT   0x20000000  /* indicates alternative ABI  */
+#endif
+#ifndef AUDIT_ARCH_MIPS64_N32
+#define AUDIT_ARCH_MIPS64_N32  (EM_MIPS|__AUDIT_ARCH_64BIT|__AUDIT_ARCH_ALT|)
```
feckert, line 51: The last `|` throws an error on my toolchain (openwrt master)

feckert:
@RayWang10 I am working on integrating of audit into my openwrt image.
OpenWrt already provides [audit version 2.8.5](https://github.com/openwrt/packages/blob/master/utils/audit/Makefile).

I have a lantiq MIPS and stumbled over your pullrequest.
I did not realized that audit only works for some architectures!
I have not looked into it further.
The compilation worked on openwrt without your patch, so i thought it will work?

But now I get the following output on my MIPS `Error sending add rule data request (Invalid argument)` on auditctl rules add comand.
My other target x86_64 with the same source base works as expected.

I thought your arch patch will solve my issue.
I have applied your patch with some changes to apply clean.
But nothing changed.
Same error.

Is this error now arch error or not?
Can someone give me a hint.

@tpetazzoni If audit works only on some architectures, we need to restrict compilation to the supported ones.
Regardless of whether the kernel supports audit, because it looks like auditd only works for some architectures.

feckert:
@stevegrubb @RayWang10
I have a MIPS arch and would like to use auditd userspace tool.
It looks like I need this pullrequest to make the userspace tool work for this arch.
What do I have to do to continue here?
How cann I help?
What else do I need?

The link `https://www.linux-mips.org/archives/linux-mips/2018-12/msg00193.html` does not exist anymore.
I am using and developing with kernel 5.4 from openwrt master branch

stevegrubb:
I am reluctant to add more architectures to userspace audit. They main reason is it takes a lot of time updating the syscall tables and no one really helps. I got stuck taking care of the Alpha processor and IA64 after a promise of help and then they disappeared. If there were more contributors to the audit subsystem, it might be another story.

PR #129, opened 2020-05-30
Add prelude SIEM support

ToToL:
Hello

Few years before, Auditd has the Prelude SIEM support.

Nowadays, Prelude SIEM is active and people ask the Prelude Team when the support of Prelude will go back inside auditd.

Here is the commit and it is fully working.

Please take a look and add it.

Thanks

Regards


stevegrubb:
I do not want to support this code. It was deleted for a reason. Its old and has not been maintained wrt new events and capabilities. Its best to give it a new home in some other repository. Or maybe merge it with prelude-lml.

ToToL:
Hello

Thanks for your feedback.

If the problem is the support of this code, I can do this. I already do it for other security probes (suricata, ossec, kismet, etc.)

I understand what you said about the age of this code but if you point me what need to be updated, I can do the job.

Here is the master repository for auditd. I don't think its a good idea to fork and push the code, people will not be aware of this possibility.

Also, I don't think Prelude-lml is a good idea because the aim of putting libprelude inside probes is to get rid of prelude-lml rules.

If it is possible to go forward on this, let me know, I'm open :)

Regards

Thomas

stevegrubb:
Going to close this out. Someone can create a standalone plugin. It doesn't have to be shipped as part of the audit project.

PR #135, opened 2020-08-27
Add a USER_MAC_STATUS audit type.

pebenito:
This is the userspace daemon equivalent to MAC_STATUS, similar to how
AVC->USER_AVC and MAC_POLICY_LOAD->USER_MAC_POLICY_LOAD.

This will allow SELinux userspace object managers to report their enforcing
changes.

Signed-off-by: Chris PeBenito <chpebeni@linux.microsoft.com>


stevegrubb:
Which daemon, if any, can change its enforcement? We're having massive trouble with userspace just reporting USER_MAC_POLICY_LOAD. Dbus, for example, gets it completely wrong and sends an AVC event.

pebenito:
I'm aware of the USER_MAC_POLICY_LOAD dbus issue. That is what has set me off on this auditing investigation. There is an underlying libselinux issue that pretty much makes it impossible for daemons to audit that. I've sent [a patch](https://lore.kernel.org/selinux/20200827125839.79006-1-chpebeni@linux.microsoft.com/T/#u), and I've already submitted a first step PR bus1/dbus-broker#238 to lay the groundwork to address this issue.

I plan to go through the other users of this libselinux code to clean up the auditing too.

XSELinux can set its enforcing mode independently from the kernel. The libselinux userspace supports this so other daemons could do this in the future. The libselinux code also will emit a message about the enforcing change which dbus will also audit as USER_AVC (as do most other users of this code).

pebenito:
A further note, the daemons that don't independently change their enforcement still change their enforcement in sync with the kernel's enforcing mode.

Issue #86, opened 2019-03-19
BUG: records from one event not grouped together


rgbriggs:
In recent testing of events for [ghak90](https://github.com/linux-audit/audit-kernel/issues/90) and [ghak110](https://github.com/linux-audit/audit-kernel/issues/110) it has been noticed that ausearch isn't grouping all the records of one event together even though the date/time/serial stamps are identical.  It may be related to the fact that both records are in the 1000 block.

ghak90:
```----
type=CONTAINER_OP msg=audit(2019-03-19 15:40:06.141:235) : op=set opid=698 old-contid=18446744073709551615 contid=123455 pid=698 uid=root auid=root tty=ttyS0 ses=1 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 comm=test.sh exe=/usr/bin/bash res=yes 
----
type=CONTAINER_ID msg=audit(2019-03-19 15:40:06.141:235) : contid=123455 
type=PROCTITLE msg=audit(2019-03-19 15:40:06.141:235) : proctitle=/bin/sh ./test.sh 
type=SYSCALL msg=audit(2019-03-19 15:40:06.141:235) : arch=x86_64 syscall=write success=yes exit=7 a0=0x1 a1=0x55e34362ceb0 a2=0x7 a3=0xa items=0 ppid=676 pid=698 auid=root uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=ttyS0 ses=1 comm=test.sh exe=/usr/bin/bash subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 key=(null) 
----
```

ghak110:
```----
time->Tue Mar 19 12:23:15 2019
type=LOGIN msg=audit(1553012595.401:219): pid=647 uid=0 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 old-auid=4294967295 auid=0 tty=ttyS0 old-ses=4294967295 ses=1 res=1
----
time->Tue Mar 19 12:23:15 2019
type=PROCTITLE msg=audit(1553012595.401:219): proctitle=2F62696E2F6C6F67696E002D70002D2D0020202020
[...]

PR #141, opened 2020-12-16
allow LOGIN event record to be grouped with its SYSCALL record

rgbriggs:
LOGIN records were not grouped with the rest of their event, records with the
identical timestamp and serial number:

----
time->Tue Mar 19 12:23:15 2019
type=LOGIN msg=audit(1553012595.401:219): pid=647 uid=0 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 old-auid=4294967295 auid=0 tty=ttyS0 old-ses=4294967295 ses=1 res=1
----
time->Tue Mar 19 12:23:15 2019
type=PROCTITLE msg=audit(1553012595.401:219): proctitle=2F62696E2F6C6F67696E002D70002D2D0020202020
type=SYSCALL msg=audit(1553012595.401:219): arch=c000003e syscall=1 success=yes exit=1 a0=3 a1=7fff3fb13380 a2=1 a3=0 items=0 ppid=1 pid=647 auid=0 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=ttyS0 ses=1 comm="login" exe="/usr/bin/login" subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 key=(null)
----

Tidy up the list/ranges of records to be in ascending order for easier parsing
and updating by humans.

See the upstream issues
	https://github.com/linux-audit/audit-userspace/issues/86

Signed-off-by: Richard Guy Briggs <rgb@redhat.com>

### DIFF
--- a/src/ausearch-lol.c
+++ b/src/ausearch-lol.c
@@ -246,12 +246,14 @@ static void check_events(lol *lo, time_t sec)
 			if (cur->l->e.sec + 2 <= sec) { 
 				cur->status = L_COMPLETE;
 				ready++;
-			} else if (cur->l->e.type == AUDIT_PROCTITLE ||
-				    cur->l->e.type < AUDIT_FIRST_EVENT ||
-				    cur->l->e.type >= AUDIT_FIRST_ANOM_MSG ||
-				    cur->l->e.type == AUDIT_KERNEL ||
-				    (cur->l->e.type >= AUDIT_MAC_UNLBL_ALLOW &&
-				    cur->l->e.type <= AUDIT_MAC_CALIPSO_DEL)) {
+			} else if ( cur->l->e.type <  AUDIT_LOGIN            ||
+				   (cur->l->e.type >  AUDIT_LOGIN           &&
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

stevegrubb:
@rgbriggs is this patch current with all the changes up to the 6.5 kernel? I think it also misses similar logic in the normalizer and auparse. Maybe this could be turned into a hidden function in auparse that can be used internally but not exposed to the outside world?

Would like to merge this for the soon to be released audit-4.0.

stevegrubb:
This patch had conflicts and I need to hurry this up. I used the second detection but the patch unnecessarily changed the order of comparisons. This logic is moved to common/common.c.

rgbriggs:
There was no hurry for 3 years.  why the 3-day rush now over a weekend?

stevegrubb:
Because this is a behavior change that should go into a major release. There is a major release in the works and this is one of the last two items that were not completed. I need to finish the major release asap because I no longer work on security. I need to wrap up the release in the next couple of days. There weren't any replies, so I fixed it myself.

PR #142, opened 2020-12-16
Ghak25 netfilter cfg.v6

rgbriggs:


stevegrubb:
I don't particularly like the function name parse_yaasao. I'd rather just call it parse_listener. While it also does netfliter, it at least is more meaningful than an abstract name.

rgbriggs:
On 2020-12-16 14:09, Steve Grubb wrote:
> I don't particularly like the function name parse_yaasao. I'd rather just call it parse_listener. While it also does netfliter, it at least is more meaningful than an abstract name. 

I didn't like it either.  I'm willing to change it.  :-)

PR #145, opened 2020-12-18
libaudit: use UINT_MAX macro in place of magic numbers

rgbriggs:
Signed-off-by: Richard Guy Briggs <rgb@redhat.com>


stevegrubb:
Not exactly a fan of this. The issue is that the kernel should define this. Userspace should use it. And this patch misses all the places where we have a uint32_t that is -1.

stevegrubb:
Closing this out. Richard, thanks for the patch. If the kernel defines a macro and uses it internally, we can use it in user space.

PR #149, opened 2021-01-11
lib/arm_table.h: update arm syscall table

yizhao1:
Refer to Glibc 2.32, add *_time64 syscalls.

Signed-off-by: Yi Zhao <yi.zhao@windriver.com>


stevegrubb:
Thanks for the patch!

Issue #148, opened 2021-01-10
BUG: Ausearch fails to correctly parse an event whose timestamp appears more than 2 seconds later than co-located events


burnalting:
Ausearch fails to correctly parse the event with id 44609 in the except from an audit.log file below. 

```
type=SYSCALL msg=audit(1609519900.161:44607): arch=c000003e syscall=59 success=yes exit=0 a0=1130ce0 a1=1125fd0 a2=1124650 a3=7ffdb79a1440 items=2 ppid=27828 pid=27856 auid=0 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=269 comm="hostname" exe="/usr/bin/hostname" subj=system_u:system_r:system_cronjob_t:s0-s0:c0.c1023 key="cmds"
type=EXECVE msg=audit(1609519900.161:44607): argc=1 a0="hostname"
type=CWD msg=audit(1609519900.161:44607):  cwd="/root"
type=PATH msg=audit(1609519900.161:44607): item=0 name="/usr/bin/hostname" inode=50333283 dev=fd:00 mode=0100755 ouid=0 ogid=0 rdev=00:00 obj=system_u:object_r:hostname_exec_t:s0 objtype=NORMAL cap_fp=0000000000000000 cap_fi=0000000000000000 cap_fe=0 cap_fver=0
type=PATH msg=audit(1609519900.161:44607): item=1 name="/lib64/ld-linux-x86-64.so.2" inode=209023 dev=fd:00 mode=0100755 ouid=0 ogid=0 rdev=00:00 obj=system_u:object_r:ld_so_t:s0 objtype=NORMAL cap_fp=0000000000000000 cap_fi=0000000000000000 cap_fe=0 cap_fver=0
type=PROCTITLE msg=audit(1609519900.161:44607): proctitle="hostname"
type=SYSCALL msg=audit(1609519896.829:44609): arch=c000003e syscall=59 success=yes exit=0 a0=254c440 a1=2527e90 a2=252b650 a3=7ffdfdce0cd0 items=2 ppid=27755 pid=27854 auid=0 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=268 comm="find" exe="/usr/bin/find" subj=system_u:system_r:system_cronjob_t:s0-s0:c0.c1023 key="cmds"
[...]

PR #150, opened 2021-01-27
Code to address issue148 providing both auditd.conf and userspace sol…

burnalting:
This PR addresses issue #148 by providing

- a configuration item in /etc/audit/auditd.conf
- user space changes by providing a --eoe-timeout option to ausearch and aureport and also a auparse_set_eoe_timeout() routine to the auparse stable of functions.

### DIFF
--- a/auparse/auditd-config.c
+++ b/auparse/auditd-config.c
@@ -359,8 +359,7 @@ static int eoe_timeout_parser(auparse_state_t *au, const char *val, int line,
 
 void free_config(struct daemon_conf *config)
 {
-	if (config->log_file != NULL)
-		free((void*)config->log_file);
+	free((void*)config->log_file);
 	config->log_file = NULL;
 }
 
--- a/auparse/auditd-config.c
+++ b/auparse/auditd-config.c
@@ -359,7 +359,7 @@ static int eoe_timeout_parser(auparse_state_t *au, const char *val, int line,
 
 void free_config(struct daemon_conf *config)
 {
-	if (config->log_file == NULL)
+	if (config->log_file != NULL)
 		free((void*)config->log_file);
 	config->log_file = NULL;
 }
--- a/auparse/auditd-config.c
+++ b/auparse/auditd-config.c
@@ -61,10 +61,13 @@ static int nv_split(char *buf, struct _pair *nv);
 static const struct kw_pair *kw_lookup(const char *val);
 static int log_file_parser(auparse_state_t *au, const char *val, int line, 
 		struct daemon_conf *config);
+static int eoe_timeout_parser(auparse_state_t *au, const char *val, int line,
+		struct daemon_conf *config);
 
 static const struct kw_pair keywords[] = 
 {
   {"log_file",		log_file_parser },
+  {"end_of_event_timeout", eoe_timeout_parser },
   { NULL,		NULL }
 };
 
@@ -100,6 +103,7 @@ void clear_config(struct daemon_conf *config)
 	config->disk_full_exe = NULL;
 	config->disk_error_action = FA_SYSLOG;
 	config->disk_error_exe = NULL;
+        config->end_of_event_timeout = EOE_TIMEOUT;
 }
 
 int aup_load_config(auparse_state_t *au, struct daemon_conf *config,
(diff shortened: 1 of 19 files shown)


auparse/auditd-config.c
```diff
@@ -323,8 +327,40 @@ static int log_file_parser(auparse_state_t *au, const char *val, int line,
...
+		return 1;
+	}
+	config->end_of_event_timeout = i;
+	return 0;
+}
+
 void free_config(struct daemon_conf *config)
 {
-	free((void*)config->log_file);
+	if (config->log_file == NULL)
```
stevegrubb, line 362: free() is advertised as just returning if NULL is passed so that we do not need to check if its NULL before calling free. Besides, I think you meant !=, but its totally unnecessary.

auparse/auparse.c
```diff
@@ -398,6 +400,32 @@ void print_lol(char *label, au_lol *lol)
...
+
+        /* Load config so we know where logs are */
+	set_aumessage_mode(au, MSG_STDERR, DBG_NO);
+	aup_load_config(au, &config, TEST_SEARCH);
+
+	// Now over-ride the interal user space configuration items
+
+	if (config.end_of_event_timeout != EOE_TIMEOUT)
+		eoe_timeout = (time_t)config.end_of_event_timeout;
+
```
stevegrubb, line 425: I'm wondering if config.log_file ought to be kept so that it can be given to setup_log_file_array() if needed? Also, the if really isn't necessary since if they are the same it's harmless.

src/auditd-config.c
```diff
@@ -1961,7 +1996,9 @@ void free_config(struct daemon_conf *config)
         free((void *)config->krb5_key_file);
 	free((void *)config->plugin_dir);
         free((void *)config_dir);
-        free(config_file);
+        if (config_file != NULL)
+		free(config_file);
```
stevegrubb, line 2000: Same comment here, free can handle NULL values.

src/ausearch-userspace.c
```diff
@@ -0,0 +1,40 @@
...
+	extern time_t   eoe_timeout;
+
+	// Load the configuration file 
+	(void)load_config(&config, TEST_SEARCH);
+
+	// Now over-ride the interal user space configuration items
+
+	if (config.end_of_event_timeout != EOE_TIMEOUT)
+		eoe_timeout = (time_t)config.end_of_event_timeout;
+
```
stevegrubb, line 38: Same comment here, the if is harmless

src/ausearch-lol.c
```diff
@@ -35,6 +35,10 @@
 static int ready = 0;
 event very_first_event;
 
+// End of Event timeout value (in seconds). This is over-ridden by auditd-conf's end_of_event_timeout configuration item
+time_t eoe_timeout = EOE_TIMEOUT;
```
stevegrubb, line 39: Should this be a static variable and we have a set function that ausearch/report calls unconditionally to initialize this with either the config file value, the passed value, or the default value?

auparse/auditd-config.c
burnalting: Correct, my bad. I did mean !=.

That said, in ausearch/aureport I was seeing the error attempting to double free memory, so wanted to remove the error message. If we don't do this, then every execution will do a double free, for we allocate this structure twice, meaning free_config() is called twice.

I suppose I was just being consistent between ausearch's free_config() and auparse's. Technically, the auparse code doesn't need this 'error message protection', but this is just in case of the sequence

```
struct daemon_conf config;
au_load_config(au, &config, ..);
do stuff;
free_config(&config);
au_load_config(au, &config, ..);
do stuff
free_config(&config);
```

src/auditd-config.c
burnalting: As before, I like to control memory management routines.

src/ausearch-lol.c
burnalting: If this is the philosophy wanted, then I can do it. Please advise.

auparse/auparse.c
burnalting: I thought about only having one struct daemon_conf structure the same way that ausearch/aureport does, but decided on minimal change ... that would have been a bit more change given we only call setup_log_file_array for only one of a number of ausource_t types.

And yes, we can ditch the if()

auparse/auditd-config.c
stevegrubb: yes for deleting the 'if'. Setting to NULL is fine

src/ausearch-lol.c
stevegrubb: I think it would be better to have the source of truth in one place and it gets set from outside by the app which should consider which of the three the values it is and make it so.

auparse/auditd-config.c
burnalting: Done

auparse/auparse.c
burnalting: The if has been removed

src/auditd-config.c
burnalting: The if has been removed.

src/ausearch-lol.c
burnalting: Done.

stevegrubb:
Looks good. Thanks!

PR #151, opened 2021-01-28
Reduce permissions with systemd

bmwiedemann:
Reduce permissions with systemd
to reduce security impact of an exploit.
This is tested with the default config on openSUSE Tumbleweed and Leap 15.2
and might need adaptation for other OSes, remote logging or container handling.

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -29,10 +29,28 @@ ExecStartPost=-/sbin/augenrules --load
 #ExecStopPost=/sbin/auditctl -R /etc/audit/audit-stop.rules
 
 ### Security Settings ###
-MemoryDenyWriteExecute=true
+CapabilityBoundingSet=CAP_AUDIT_CONTROL CAP_AUDIT_READ CAP_SYS_NICE CAP_SYS_RESOURCE
 LockPersonality=true
+MemoryDenyWriteExecute=true
+PrivateDevices=true
+PrivateNetwork=true
+PrivateUsers=no
+ProtectClock=true
 ProtectControlGroups=true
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for the patch. I am afraid to take the whole patch as it would likely affect child processes that may need different permissions than auditd. There are parts of the patch that I think are good. I'm not sure how to proceed. Should I cherrypick the setting that are universally safe?

 And also, the upstream audit project is neutral on what OS it's deployed to. I, for example, run audit-3.0.1 on RHEL 7 which uses systemd-219 which does not have advanced security options. I suspect other distributions may also have an older systemd. So, how do we know which options are safe for any random deployment?

bmwiedemann:
Feel free to cherry pick. I know way too little about auditd to make good decisions.

With Debian 10's systemd-241 I saw in the journal:
> Unknown lvalue 'RestrictSUIDSGID' in section 'Service', ignoring

if older versions do the same, it should be safe. Could just confuse or annoy users.

Maryse47:
@stevegrubb do you still plan to cherry-pick some of the options? As it was said above older systemd instances will just ignore unsupported ones.

stevegrubb:
After checking this list with older versions of systemd and considering what future and in-development plugins might need, I have added 2 new restrictions. The main issue really is not knowing what plugins may need and causing a nasty surprise or unexpected debug trouble. Auditd and its plugins should be contained by MAC policy which is a bit more flexible in catering to the needs of plugins.

stevegrubb:
I'm going to close this out. Thanks for the suggestion!

Issue #152, opened 2021-02-05
Multiple definition of free_config and clear_config when static building


hube12:
Hello,
There is still currently some build errors when building the static library, it also worth noting that since #147 the current bug in 3.0 was fixed thus the only way to build a static lib is to either apply that patch or build the current branch.

```sh
/usr/bin/ld:  /audit-userspace/linux-audit-audit-userspace-0771ea4/auparse/auditd-config.c:82: multiple definition of `clear_config'; 
auditd-auditd-config.o:/audit-userspace/linux-audit-audit-userspace-0771ea4/src/auditd-config.c:312: first defined here

/usr/bin/ld: /audit-userspace/linux-audit-audit-userspace-0771ea4/auparse/auditd-config.c:361: multiple definition of `free_config'; 
auditd-auditd-config.o:/audit-userspace/linux-audit-audit-userspace-0771ea4/src/auditd-config.c:1986: first defined here
```

stevegrubb:
I renamed a couple auparse functions and checked them in. Hopefully, this fixes the issue.

hube12:
Hi, thanks for your work I opened #153 with the rest of the changes, have a wonderful day.

stevegrubb:
Hmm. The pull request is quite a bit more than I thought it would be. Thanks. There are a couple that I might want to rename. I might download the patch, doctor it up and apply manually. If I do that right, it should still link back to you. That has been my only hesitation in accepting the patch.

hube12:
Hi, I totally get it, I am happy either way, I don't see that as a big changes, more a fix to allow a edge case (static compilation), thanks for taking that into account !

stevegrubb:
The patch has been applied with modifications. Thanks for sending it. If you notice any problems that it should have covered, please feel free to re-open this issue.

PR #153, opened 2021-02-08
Rename functions to allow static compilation

hube12:
Hello,
I saw that you made some changes after #152  and I am really grateful for those, however some more were needed (this PR makes those naming changes), I also modified in my branch https://github.com/hube12/audit-userspace/tree/static the makefile to make static bindings for python:
My commands I ran (if needed I can provide a python configuration script that do it automatically):
```sh
git clone https://github.com/hube12/audit-userspace
cd audit-userspace
# since I need python bindings statically compiled I use this branch with changed makefiles
git checkout static
./autogen.sh
./configure --with-python3=yes --without-golang --with-debug=yes --with-warn=yes --with-arm=yes --with-aarch64=yes --with-apparmor=yes --with-libwrap=no --with-libcap-ng=no --with-pic=no --enable-systemd=no --enable-static=yes --enable-shared=no --enable-gssapi-krb5=no --enable-zos-remote=no --enable-listener=no --enable-fast-install=no --enable-libtool-lock=no --enable-dependency-tracking=no --enable-silent-rules=no --prefix=$PWD/audit_static_lib
make -j8
make install
```

Thanks for your time and I would totally understand if you want to rename them differently, I only added prefix to match up to the filename but I do not know as well as you libaudit.

Cheers,
Neil


stevegrubb:
OK, looking over the changes...I think I want to leave the function named interpret the way it is in auparse. Renaming the one in ausearch-report should be enough to avoid the collision. I applied the patch manually. Thanks.

ajs124:
This still seems to be a problem, because there's also a global variable named `interpret`.
```
/build/audit-3.0.1/auparse/interpret.c:2860: multiple definition of `interpret'; auditctl-auditctl.o:/build/audit-3.0.1/src/auditctl.c:60: first defined here
```

stevegrubb:
commit 0c2ae97 should fix this last issue.

PR #159, opened 2021-03-27
Avoid Missing return statement in value returning function warning

dirkmueller:
This isn't actually using the declared return value anywhere.

### DIFF
--- a/src/libev/ev_iouring.c
+++ b/src/libev/ev_iouring.c
@@ -287,7 +287,7 @@ iouring_sqe_get (EV_P)
 }
 
 inline_size
-struct io_uring_sqe *
+void
 iouring_sqe_submit (EV_P_ struct io_uring_sqe *sqe)
 {
   unsigned idx = sqe - EV_SQES;


stevegrubb:
Thanks for the patch. This is embedded code. You might want to send the patch to libev upstream. But this is the tip of the io_ring iceberg. There is much more badness if you run this through static analysis. The io_ring backend is not used by the audit system. We use either select or epoll depending on if it's listening for connections.

stevegrubb:
I am going to close this out. I appreciate your taking the time to send this. Auditd by default uses the select backend. There is no way to specify the io_uring backend after compilation. While the warning shows up in the build log, auditd is not affected. Hopefully the next release of libev has this and other io_uring problems cleaned up.

dirkmueller:
Thanks for taking the time to explain the context. The patch went upstream (http://cvs.schmorp.de/libev/ev_iouring.c?r1=1.24&r2=1.25) and hopefully is in the next release. I hope that you are going to update when it becomes available. for now I need this change to bypass a local linting check.

PR #170, opened 2021-04-15
Add man-page set_aumessage_mode(3)

miztake:
So far, the man-page of set_message_mode(3) has existed, but the man-page of set_aumessage_mode(3) has not.
set_message_mode does not exist in libaudit.
Therefore, the man-page of set_message_mode(3) was deleted and the man-page of set_aumessage_mode(3) was added.


stevegrubb:
Thanks for all the patches!

PR #171, opened 2021-04-19
Fix the default location for zos-remote.conf

yixiangzhike:
All audispd config files have been moved to "/etc/audit" in the commit a5f7baf0917a60cacbd237ba1f1512db63445031.
 The file zos-remote.conf is still configured under /etc/audisp in audispd-zos-remote.conf.


stevegrubb:
Thanks for the patch

PR #178, opened 2021-04-22
Fix not to list own man-page in "SEE ALSO"

miztake:
There is a man-page that references to its own man-page.
Fixed not to refer to own man-page.


stevegrubb:
Thanks!

PR #180, opened 2021-04-22
Fix the title in the Title line according to man-pages(7)

miztake:


stevegrubb:
Thanks for all the patches!

PR #197, opened 2021-04-27
Add man-page for audit_value_needs_encoding()

miztake:


stevegrubb:
Thanks for the patch. Makefile.am may need touching up, but I'll take care of that.

PR #198, opened 2021-04-28
[RFC] syscall: improve syscall sync with update script

feckert:
@stevegrubb
**I would like to mention right away that this is a proof of concept.
I have not checked the changed syscall tables and I have not compiled them!**

The syscall tables for the individual architectures were generated from the syscall tables of this [project](https://github.com/hrw/syscalls-table.git)

I used the following new command `syscall-table-generate` in the the contrib dir to generate the` x86_64_table.h` file:
```
cd <audit-userspace-project-dir>/contrib
./syscall-table-generate <syscall-table-project-dir>/tables/syscalls-x86_64 ../lib/x86_64_table.h
```


stevegrubb:
Interesting idea. There seems to be a bunch that leave and some new ones show up. The syscall rules are sensitive to name changes which do things like add "64".

feckert:
> The syscall rules are sensitive to name changes which do things like add "64".

I cannot judge whether there will be a problem. I know too little about audit framework. I am willing to deal more with the internes. It should be checked whether these are the official names and so this one should be used.

In order to be able to judge all this better, I have a few more questions for you.
It would be great if you could answer them.

1. I use and work with OpenWrt and there is the auditd for the MIPS architecture, and it also runs on my target. Does this make sense at all if the syscall table for this architecture does not exist?

2. My usecase is, I want to watch a file change on my target. For x86_64 it works as expect If I execute the following command.
`root@test /etc/audit # auditctl -w /etc/passwd -p r`, but if I execute the command on my MIPS target I get the following output.
`Error sending add rule data request (Invalid argument)`. I thing this is because the syscall table is not supprted in audit-userspace.

3. There is an effort in Openwrt to use selinux, that is as far as I know also the audit involved. Does that work at all on architectures where there is no userland support because the syscall tables are missing?

Sorry if I ask stupid questions, but I can't find anything about this topic on the internet. The audit and selinux are mainly used on server systems. That is why it works there without problems. The syscall tables for this architectures are included in the auditd userland project.

feckert:
@stevegrubb I have applied the [pullrequest](https://github.com/linux-audit/audit-userspace/pull/115) on my openwrt build with some changes to apply corectly.

If I do not add the pullrequest in my build, I get the following output on my mips target:
```
root@devel ausyscall --dump
Unable to detect machine type

root@devel ~ # auditctl -s
enabled 1
failure 1
pid 13416
rate_limit 0
backlog_limit 64
lost 0
backlog 0
backlog_wait_time 6000
loginuid_immutable 0 unlocked

root@devel ~ # auditctl -w /etc/shadow -p wa
Error sending add rule data request (Invalid argument)
```

If I then execute the same commands with the syscall table change from the [pullrequest](https://github.com/linux-audit/audit-userspace/pull/115) for mips, I get the following outputs:

```
root@devel ~ # ausyscall --dump
Using mips syscall table:
4000    syscall
4001    exit
4002    fork
4003    read
4004    write
4005    open
4006    close
4007    waitpid
4008    creat
4009    link
4010    unlink
4011    execve
4012    chdir
<and so on>

root@devel ~ # auditctl -s
enabled 1
failure 1
pid 13956
rate_limit 0
backlog_limit 64
lost 0
backlog 0
backlog_wait_time 6000
loginuid_immutable 0 unlocked

root@devel ~ # auditctl -w /etc/shadow -p wa
Error sending add rule data request (Invalid argument)
```

What is strange now is that I get the same error with the last command when I want to add a rule.
I would now have thought that if I added the syscall table for mips build, then that would now work.
I don't know what else I can do.
Any help would be great.

feckert:
@stevegrubb Can't you pleasee give me a hint what I can?

feckert:
I'm closing the pullrequest, otherwise it will just rot here

stevegrubb:
The error message comes from the kernel. I don't know if you are using a kernel with all the audit pieces enabled. But I'd look at that and use strace to dump all of the sendto data that goes into the kernel. Might also need to puts some printk's in the kernel to see what it doesn't like.

PR #199, opened 2021-05-01
Add man-page for audit_encode_value()

miztake:


stevegrubb:
Thanks for the patch!

PR #201, opened 2021-05-04
Fix "sequence" of static variable used in audit_send() to be mutually…

miztake:
… exclusive

### DIFF
--- a/lib/netlink.c
+++ b/lib/netlink.c
@@ -28,12 +28,15 @@
 #include <fcntl.h>
 #include <time.h>
 #include <sys/poll.h>
+#include <pthread.h>
 #include "libaudit.h"
 #include "private.h"
 
 #ifndef NETLINK_AUDIT
 #define NETLINK_AUDIT 9
 #endif
+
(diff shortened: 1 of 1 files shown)


stevegrubb:
Hmm. I suppose that it is theoretically possible to have an update collision to sequence, in practice there should not be a problem. This can only happen with 2 prerequisites: the program is multi-threaded, the program sends lots of events. No program should be sending lots of events. Libvirtd might be the only multi-threaded program that I know of that sends audit events.

I am wondering if a lower overhead approach might be better? Gcc supports atomic_int in stdatomic.h. Since this is just incrementing a counter, I wonder if that might be better? Or maybe __atomic_fetch_add(&sequence, 1, __ATOMIC_SEQ_CST); ?

My only concern is pthread_lock/unlock is slower for a case that really shouldn't happen.

miztake:
Hi, @stevegrubb

Thank you for your comment.

I understand that adding "slow" processing for less common problems is not desirable.
I would reconsider using stdatomic.h instead of pthread_lock/unlock.
Therefore, I will close this PR.

Thank you for your cooperation.

PR #214, opened 2021-09-06
Parse comma separated option list in check_rule_mismatch

Blarse:


Blarse:
Hi, 

Without these changes the entire comma separated list was passed to audit_rule_syscallbyname_data, instead of separate syscalls.

stevegrubb:
Is there a reproducer that I could test with?

Blarse:
> Is there a reproducer that I could test with?

Yes, to reproduce this you need syscalls to be the same for 32b and 64b.
The problem initially was discovered on ppc architecture. 

With comma separated list auditctl gave a warning:
> \# auditctl -a  always,exit -S open,openat -F exit=-EPERM -k open-eperm
> WARNING - 32/64 bit syscall mismatch, you should specify an arch

But with a separate options it didn't:
> \# auditctl -a  always,exit -S open -S openat -F exit=-EPERM -k open-eperm

Generally speaking you can see the wrong behavior in gdb on any architecture.
parse_syscall() internally splits comma separated list provided by optarg, but after that check_rule_mismatch gets the whole list and therefore throws a warning when audit_rule_syscallbyname_data could not lookup a syscall named as a list.

Writing this I just thought why don't we move check_rule_mismatch to parse_syscall, so we don't need to parse optarg twice?

stevegrubb:
OK, I see what you are talking about.

stevegrubb:
After reviewing and testing the patch, I see one thing that should probably be fixed. The memcmp should come after it finishes strtok.IOW, it needs to be after the loop and before the free. And the memset should be done once before entering the loop probably before the first strtok call. The syscalls are in a large bitmask. The intention is to completely walk the list of syscalls adding each to the bitmask. Then afterwards, compare the two bitmask to see if they are the same.

I don't know how this is a PPC only problem. You only go into this loop if it can't determine the arch.

stevegrubb:
Let me know what your preference is: 1) updating the patch to move memset & memcmp outside of the loop, or 2) I accept "as is" and fix it in a subsequent commit. It doesn't matter to me which way we do it. The core idea is good.

Blarse:
> You only go into this loop if it can't determine the arch.

Well, it determines and sets `_audit_elf` on the first run, but updates `unknown_arch` only on the second.
Is this intended behavior?

> I don't know how this is a PPC only problem. 

This is PPC only problem because ppc has the same syscall codes for b32 and b64.
On the other systems this warning is expected.

PR #216, opened 2021-09-08
ausyscall: Add support for 'b32' and 'b64' aliases

Blarse:


stevegrubb:
This also needs a change in the man page. I'll take care of it.

Issue #211, opened 2021-08-21
Automatically restart the auditd service when the service is terminated unexpectedly.


StevenYGui:
Is it possible to add a line "Restart=on-failure" in file "init.d/auditd.service", then let systemd to automatically restart the auditd service when the service is terminated unexpectedly?
Auditd is an important service in linux, so it is better if it keeping actively.

stevegrubb:
The main reason that is not done is because auditd halts when it finds a problem in its config file. It it were automatically restarted, that would be unhelpful. I'm also not sure of the interaction between systemd and the service command. If you kill auditd with the service stop command and systemd restarts it...that also is not helpful.

StevenYGui:
yes, it can not restart auditd if stopped by service command, but it is useful for "killing -9", maybe can avoid misoperations or simple attacks.

travier:
As per [systemd.service (Restart=)](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=):

```
When the death of the process is a result of systemd operation (e.g. service stop or restart), the service will not be restarted.
```

And also [systemd.service (RestartPreventExitStatus=)](https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartPreventExitStatus=):

```
Takes a list of exit status definitions that, when returned by the main service process, will prevent automatic service restarts, regardless of the restart setting configured with Restart=. 
```

We could have:

```
[Service]
Restart=on-failure
RestartPreventExitStatus=<wrong config exit code here>
```

travier:
From https://github.com/linux-audit/audit-userspace/blob/master/src/auditd.c#L714, it's `6`. Making a PR.

PR #217, opened 2021-09-29
auditd.service: Restart 'on-failure', ignoring some exit codes

travier:
Use `Restart=on-failure` to automatically restart `auditd`. Do not restart for intentional exits. See EXIT CODES section in auditd(8).

See:
- https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=
- https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartPreventExitStatus=

Fixes: https://github.com/linux-audit/audit-userspace/issues/211

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -27,6 +27,9 @@ ExecStartPost=-/sbin/augenrules --load
 # By default we don't clear the rules on exit. To enable this, uncomment
 # the next line after copying the file to /etc/systemd/system/auditd.service
 #ExecStopPost=/sbin/auditctl -R /etc/audit/audit-stop.rules
+Restart=on-failure
+# Do not restart for intentional exits. See EXIT CODES section in auditd(8).
+RestartPreventExitStatus=2 4 6
 
 ### Security Settings ###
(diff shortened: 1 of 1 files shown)


travier:
Exit code from https://github.com/linux-audit/audit-userspace/blob/master/src/auditd.c#L714

stevegrubb:
I did some testing and it seems to not cause problems. But I think exit code 2 or 4 will also send it into a loop,

travier:
Sure. Exit code 4 is https://github.com/linux-audit/audit-userspace/blob/master/src/auditd.c#L685. What is 2?

stevegrubb:
I updated the auditd.8 man page yesterday to discuss the exit codes. SysVinit had standard exit codes that daemons needed to obey. An exit with 2 was a command line argument problem.

travier:
For reference: https://github.com/linux-audit/audit-userspace/commit/190c40d3cb5806f84a8dcd2968042433ad59d587

Updated! Thanks

stevegrubb:
Thanks!

PR #225, opened 2021-11-02
Add man-page for audit_syscall_to_name()

miztake:


stevegrubb:
Thanks for the patch!

PR #228, opened 2021-11-29
Security hardening for auditd.service, create augenrules.service

ematsumiya:
As part of [openSUSE's security hardening effort](https://en.opensuse.org/openSUSE:Security_Features#Systemd_hardening_effort), implement some of systemd's measures to auditd.service.

Also split augenrules into a separate service so we can have an exception there for read/write to /etc/audit, since auditd.service doesn't need such access.

### DIFF
--- a/init.d/auditd.service
+++ b/init.d/auditd.service
@@ -11,22 +11,22 @@ After=local-fs.target systemd-tmpfiles-setup.service
 Before=sysinit.target shutdown.target
 ##Before=shutdown.target
 Conflicts=shutdown.target
-RefuseManualStop=yes
 ConditionKernelCommandLine=!audit=0
 Documentation=man:auditd(8) https://github.com/linux-audit/audit-documentation
+Requires=augenrules.service
+# This unit clears rules on stop, so make sure that augenrules runs again
+PropagatesStopTo=augenrules.service
 
 [Service]
 Type=forking
 PIDFile=/run/auditd.pid
 ExecStart=/sbin/auditd
-## To not use augenrules, copy this file to /etc/systemd/system/auditd.service
-## and comment/delete the next line and uncomment the auditctl line.
-## NOTE: augenrules expect any rules to be added to /etc/audit/rules.d/
-ExecStartPost=-/sbin/augenrules --load
+## To not use augenrules: copy this file to /etc/systemd/system/auditd.service,
+## uncomment the next line, and comment the Requires=augenrules.service above.
 #ExecStartPost=-/sbin/auditctl -R /etc/audit/audit.rules
-# By default we don't clear the rules on exit. To enable this, uncomment
+# By default we clear the rules on exit. To disable this, comment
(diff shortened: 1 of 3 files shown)


init.d/auditd.service
```diff
@@ -38,6 +36,12 @@ ProtectControlGroups=true
 ProtectKernelModules=true
 ProtectHome=true
 RestrictRealtime=true
+ProtectSystem=full
+ProtectHostname=true
+ProtectClock=true
+ProtectKernelTunables=true
+ProtectKernelLogs=true
+# end of automatic additions
```
Vogtinator, line 44: Leftover

init.d/auditd.service
```diff
@@ -19,10 +19,8 @@ Documentation=man:auditd(8) https://github.com/linux-audit/audit-documentation
 Type=forking
 PIDFile=/run/auditd.pid
 ExecStart=/sbin/auditd
-## To not use augenrules, copy this file to /etc/systemd/system/auditd.service
-## and comment/delete the next line and uncomment the auditctl line.
-## NOTE: augenrules expect any rules to be added to /etc/audit/rules.d/
-ExecStartPost=-/sbin/augenrules --load
+## To not use augenrules: copy this file to /etc/systemd/system/auditd.service,
```
Vogtinator, line 22: FWICT this is missing the inter-unit dependencies

stevegrubb:
Thanks for the patch. There have been others wanting to harden auditd.service. I have picked off as much as I think we can support prior to this patch. See PR #151.

However, I might be inclined to take just the auditctl service with appropriate adjustment of audit service to completely drop any mention of auditctl. If this is integrated, I would schedule it for a future major release since it would be a significant change to managing the audit subsystem.

ematsumiya:
> However, I might be inclined to take just the auditctl service with appropriate adjustment of audit service to completely drop any mention of auditctl. If this is integrated, I would schedule it for a future major release since it would be a significant change to managing the audit subsystem.

I'm not sure I understand what you mean here. Did you mean augenrules.service (instead of auditctl service)?

stevegrubb:
Yes. Sorry for the confusion.

ematsumiya:
Ok. I think we can start with that.

But the same security measures for the augenrules.service are the ones I'm proposing for auditd.service => if they break for one, they'll break for the other. And compared with PR #151, this PR has way less restrictions (i.e. less room for trouble).

FWIW these same rules (not augenrules.service) has been in openSUSE Tumbleweed audit for about 3 months now.

stevegrubb:
The issue is the plugins to the audit daemon and not knowing what they require. There are third party plugins that could be affected if the policy is too tight. On the other hand, you can lock up augenrules.service as tight as you want. It is a one shot and has no plugins. That means you can reason about what it needs.

ematsumiya:
Thanks, Steve.
I'm closing this for now. I'll reassess our approach and needs.

PR #234, opened 2022-02-03
consider max_log_file == 0 as unlimited

ematsumiya:
If config file has, for example:

max_log_file = 0 # or unset
max_log_file_action = ROTATE # or KEEP_LOGS
num_logs = 100

auditd will rotate the logs with a single event on it, sometimes even
with no events at all, creating @num_logs rotated logs on disk.

If both max_log_file and num_logs are 0 or unset, this bug doesn't
happen; a single log file is created and appended to indefinitely.

This patch makes auditd consider that when max_log_file is 0 or
unset in the config file, the log file shall grow unlimitedly, leaving
any size checks and actions for space_left/disk_full, just as the
situation mentioned right above.

Signed-off-by: Enzo Matsumiya <ematsumiya@suse.de>

### DIFF
--- a/src/auditd-event.c
+++ b/src/auditd-event.c
@@ -1091,6 +1091,15 @@ static void rotate_logs(unsigned int num_logs, unsigned int keep_logs)
 		return;
 	}
 
+	/* If max_log_file is unset (0 by default), consider it 'unlimited',
+	 * i.e. do not rotate logs, keep appending to log_file.
+	 */
+	if (config->max_log_size == 0) {
+		audit_msg(LOG_NOTICE,
+			"Log rotation disabled, max_log_file is 0 (unlimited), skipping");
+		return;
+	}
+
 	/* Close audit file. fchmod and fchown errors are not fatal because we
 	 * already adjusted log file permissions and ownership when opening the
 	 * log file. */


JAORMX (approved):

ematsumiya:
@stevegrubb got a chance to review this one? Thanks!

ematsumiya:
Ping.

stevegrubb:
Normally when we change a behavior like this, we need to also update the man page entry for the config item.

stevegrubb:
Looking through the code, don't you get the same effect with: max_log_file_action = ignore ?

ematsumiya:
@stevegrubb Yes, you're right. As stated in the commit message, this only occurs with those setting combinations.
Our customer simply removed the "max_log_file" line from the default auditd.conf and expected that there would be no logfile size checks, hence this patch.
So effectively, yes, this patch makes "max_log_file = 0" or an auditd.conf without a max_log_file entry act like "max_log_file_action = ignore".

I'll can update man page if this is approved. Thanks.

stevegrubb:
Well, I'd rather not want make 2 ways to do the same thing. It gets confusing for people. For example, we used to allow any order for the filter and action in audit rules:  -a exit,always  -a always,exit. It really didn't matter. But it confused people. They didn't understand order didn't matter. They wanted one way to do it so they know the syntax is right. So, I locked it down and confusion went away.

stevegrubb:
I think it's time to close this out. Thanks for pointing out this issue. Do you think we should amend the documentation to better highlight the max_log_file_action = ignore way to let logs grow unconditionally?

ematsumiya:
> I think it's time to close this out. Thanks for pointing out this issue. Do you think we should amend the documentation to better highlight the max_log_file_action = ignore way to let logs grow unconditionally?

That would be good, thanks. Feel free to close this.

PR #238, opened 2022-03-01
Add man-page for conversion of name and errno

miztake:


stevegrubb:
Thanks for the man pages.

PR #240, opened 2022-03-07
Add man-page for audit_name_to_fstype()

miztake:


stevegrubb:
Thanks for the patch!

PR #241, opened 2022-03-07
Fix Makefile with the addition of audit_fstype_to_name.3 and audit_na…

miztake:
…me_to_fstype.3


stevegrubb:
Thanks for the patch!

PR #246, opened 2022-03-16
Add man-page for audit_flag_to_name()

miztake:


stevegrubb:
Thanks for the patches. When you do two related pages that translate one way and then the other, they can be combined into 1 page that covers both functions. If you plan to document others, that might save time. Thanks!

miztake:
Dear @stevegrubb,

Thank you for your comment.
As you say, I would consider integrating two man-pages into one, such as followings:

  - audit_xxxx_to_name
  - audit_name_to_xxxx

Issue #252, opened 2022-03-22
audit fails to build with linux-5.17 headers


trofi:
Noticed build failure when tried to update linux-headers from `5.16` to `5.17` in downstream distribution.

Both current git checkout and older releases fail for me as:

```
audit> swig -o audit_wrap.c -python -I. -I../../.. -I../../../lib -I/<<NIX>>/python3-3.9.11/include/python3.9 ./../src/auditswig.i
audit> /<<NIX>>/bash-5.1-p16/bin/bash ../../../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../../..  -I. -I../../.. -I../../../lib -I/<<NIX>>/python3-3.9.11/include/python3.9  -shared -g -O2 -c -o _audit_la-audit_wrap.lo `test -f 'audit_wrap.c' || echo './'`audit_wrap.c
audit> libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../.. -I. -I../../.. -I../../../lib -I/<<NIX>>/python3-3.9.11/include/python3.9 -g -O2 -c audit_wrap.c  -fPIC -DPIC -o .libs/_audit_la-audit_wrap.o
audit> audit_wrap.c: In function '_wrap_audit_rule_data_buf_set':
audit> audit_wrap.c:5010:17: error: cast specifies array type
audit>  5010 |     arg1->buf = (char [])(char *)memcpy(malloc((size)*sizeof(char)), (const char *)(arg2), sizeof(char)*(size));
audit>       |                 ^
audit> audit_wrap.c:5010:15: error: invalid use of flexible array member
audit>  5010 |     arg1->buf = (char [])(char *)memcpy(malloc((size)*sizeof(char)), (const char *)(arg2), sizeof(char)*(size));
audit>       |               ^
```

I suspect it's a side-effect of https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ed98ea2128b6fd83bce13716edf8f5fe6c47f574 change:

```diff
@@ -514,1 +514,1 @@ struct audit_rule_data {
-	char		buf[0];	/* string fields buffer */
+	char		buf[];	/* string fields buffer */
```

Blarse:
Hi, I had the same problem yesterday after updating headers to 5.17.

[...]

PR #253, opened 2022-03-23
auditswig.i: avoid setter generation for audit_rule_data::buf

trofi:
As it's a flexible array generated code was never safe to use.
With kernel's https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ed98ea2128b6fd83bce13716edf8f5fe6c47f574
change it's a build failure now:

    audit> audit_wrap.c:5010:15: error: invalid use of flexible array member
    audit>  5010 |     arg1->buf = (char [])(char *)memcpy(malloc((size)*sizeof(char)), (const char *)(arg2), sizeof(char)*(size));
    audit>       |               ^

Let's avoid setter generation entirely.

Closes: https://github.com/linux-audit/audit-userspace/issues/252

### DIFF
--- a/bindings/swig/src/auditswig.i
+++ b/bindings/swig/src/auditswig.i
@@ -39,6 +39,10 @@ signed
 #define __attribute(X) /*nothing*/
 typedef unsigned __u32;
 typedef unsigned uid_t;
+/* Sidestep SWIG's limitation of handling c99 Flexible arrays by not:
+ * generating setters against them: https://github.com/swig/swig/issues/1699
+ */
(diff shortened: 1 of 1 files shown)


trofi:
Reading https://github.com/linux-audit/audit-userspace/issues/252#issuecomment-1078595249 looks like the suggestion is to use what fedora uses instead of avoiding generation of broken wrapper.

stevegrubb:
I applied this patch manually since it was closed. Thanks.

PR #259, opened 2022-05-04
Fix building with Python 3.5

mikhailnov:
Fixes: 9976d2c ("Fix the other instance of PyTypeObject which uses .tp_compare which is now deprecated")

Build error was:
```
DEBUG: ../../../bindings/python/auparse_python.c:231:6: error: 'PyTypeObject' {aka 'struct _typeobject'} has no member named 'tp_compare'<...>
DEBUG:      .tp_compare = AuEvent_compare,
DEBUG:       ^~~~~~~~~~
```


stevegrubb:
Thanks for the patch!

PR #264, opened 2022-05-27
Add definitions for messages that started appearing in Ubuntu 22.04

hillu:
These seem to have been introduced in a patchset "LSM: Module stacking
for AppArmor" (<https://lkml.org/lkml/2022/4/18/1558>) that has not
been merged upstream yet.


stevegrubb:
I was going to do an update once they are merged upstream. My recollection was that this was going into kernel-next for some testing before going to main line.

disasmwinnie:
Since the patch is in Ubuntu's kernel already, the following log lines are pretty widespread.

``type=UNKNOWN[1420] msg=audit(1659685796.323:886): subj_apparmor=unconfined``

To be honest, we have an own level of suffering with this due post-processing of the logs, which might not be relevant for the majority of Auditd users. I would understand if you want to leave it out for now. A merge into the following minor release would be greatly appreciated, though.  It think it would raise the likelyhood of Canonical update their Auditd-version quickly.

PR #269, opened 2022-08-03
Re-implement record field list as a linked list

wolneykien:
```
int nvlist_append(nvlist *l, nvnode *node)
{
	// FIXME: on overflow switch to linked list
	if (l->cnt >= NFIELDS || node->name == NULL)
	if (node->name == NULL)
		return 1;
    ...
}
```

Okay, let's switch as we have an overflow with `EXECVE` records
containing hundreds of arguments.

Signed-off-by: Paul Wolneykien <manowar@altlinux.org>

### DIFF
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -3144,8 +3144,7 @@ char *auparse_do_interpretation(int type, const idata *id,
 	if (interpretation_list_cnt()) {
 		nvlist_first(&il);
 		if (nvlist_find_name(&il, id->name)) {
-			nvnode* node = &il.array[il.cur];
-			const char *val = node->interp_val;
+			const char *val = nvlist_get_cur_val_interp(&il);
 
 			if (val) {
 				// If we don't know what it is when auditd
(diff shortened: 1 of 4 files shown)


stevegrubb:
I haven't looked at the patch yet. But I wanted to check on one thing first. The switch to the array was to speed things up. It is measurably faster. If a case was ever found where it overflowed, my intention was to offload the overflow to a linked list after consuming the whole array. Does your patch do this or revert to a linked list?

wolneykien:
Ah, you wanted it that way! No, the current patch is just a replace of the array with the linked list. OK, I'll try to rewrite it as you said: first, the array and then the linked list as a fallback on _actual_ overflow.

stevegrubb:
Thanks!

wolneykien:
Hm... What about using `realloc()` then? This way, we can stay with the array model but make it, for instance, twice bigger on each overflow.

wolneykien:
Here's the realloc patch: https://github.com/linux-audit/audit-userspace/pull/270 .

stevegrubb:
Since there is another patch fixing this via realloc, I think we can close this out.

PR #270, opened 2022-08-03
Patch: realloc on field list overflow

wolneykien:
Re-alloc on field array overflow. Not a linked list as it's slower that the array.


wolneykien:
Hi! Just added records with more than 36 fields to the test suite.

stevegrubb:
OK, I looked this over and have only one concern. (Thanks for the patch btw!) The nvlist_clear function is used to release all memory used by the nvlist. The patch reallocs the array back to NFIELDS which seems reasonable as that mimics how it would be with a static sized array. But we'll leak memory especially as used in ellist.c.

This is simple enough I can fix it after the merge. Or if you want to fix that issue I'll merge the whole thing. Up to you.

wolneykien:
Oh, thanks! It was stupid enough to mimic static behaviour with dynamic memory…
Fixed that issue and also indentation.

stevegrubb:
Thanks for the patch! Merging.

PR #271, opened 2022-08-04
auditd: avoid_oom_killer: fix typo open -> opt

bauen1:
Signed-off-by: Jonathan Hettwer <j2468h@gmail.com>


bauen1:
Actually, I've just noticed that I've misread that and the proper "Unable to adjust out of memory score" message is below that.

Sorry for the noise

PR #277, opened 2022-09-16
fix: use normal permissions for augenrules

wpyoga:
It is not necessary to restrict the permissions of `augenrules` to `0750`. If the calling user doesn't have sufficient permissions, its operation fails anyway. For security, the current restrictive permissions for the configuration files in `/etc/audit` should be enough.


stevegrubb:
This was fixed another way based on another issue that was filed.

PR #278, opened 2022-09-16
fix: error message did not mention exe type

wpyoga:
This PR is for fixing a wrong error message regarding the `exclude` filter.

If I have a rule like this:

```
-A exclude,always -F exe=/usr/bin/sudo -F success=1
```

When I run `augenctrl --load`, I get this error message:

```
Only msgtype, *uid, *gid, pid, and subj* fields can be used with exclude filter
```

However, it is not accurate. The man page says:

```
Events can be excluded  by  process  ID,  user  ID,
group  ID,  login  user  ID, message type, subject context, or exe‐
cutable name.
```

I tried this rule

```
-A exclude,always -F exe=/usr/bin/sudo
```

And it works.

### DIFF
--- a/lib/errormsg.h
+++ b/lib/errormsg.h
@@ -86,7 +86,7 @@ static const struct msg_tab err_msgtab[] = {
     { -EAU_MSGTYPEEXCLUDEUSER,	0, "msgtype field can only be used with exclude or user filter list" },
     { -EAU_UPGRADEFAIL,		0, "Failed upgrading rule" },
     { -EAU_STRTOOLONG,		0, "String value too long" },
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for pointing this out. However, it looks like this may be out of sync with the kernel. It appears the kernel now allows nearly everything on exclude filter. A much bigger fix is needed and probably revising the error messages.

wpyoga:
I actually tried to look into what the kernel accepts, not being a kernel developer myself. I went as far as [kernel/auditfilter.c](https://github.com/linux-audit/audit-kernel/blob/69bb189e81e081a4e92e8fa4cbcc9fd407252dc1/kernel/auditfilter.c), but I have to admit I don't really understand what I'm seeing...

I'll leave it to you, then :)

stevegrubb:
Commit 0b97b32 should fix this. I think there was a misplaced else.

Issue #279, opened 2022-09-16
`auditctl -R` always prepends rules instead of replacing the list of rules


wpyoga:
When `auditctl -R rulesfile` is executed, the rules inside the file is always prepended to the current list of rules. The list of rules is not replaced. `auditctl` itself returns success. An example of this usage is when `augenrules --load` is executed.

Is this intended behavior? I had thought (intuitively) that the `-R` option should in effect reload all the rules, so all the rules are removed and replaced by the ones defined in the rules file.

As you can see below:

```sh
[william@t14 ~]$ echo '-A exclude,always' > test-audit.rules
[william@t14 ~]$ echo '-A exclude,never' > test-audit-2.rules
[william@t14 ~]$ sudo chown root test-audit.rules test-audit-2.rules 
[william@t14 ~]$ sudo auditctl -D ; echo $? ; sudo auditctl -l
No rules
0
No rules
[william@t14 ~]$ sudo auditctl -R test-audit.rules ; echo $? ; sudo auditctl -l
0
-a always,exclude
[william@t14 ~]$ sudo auditctl -R test-audit.rules ; echo $? ; sudo auditctl -l
0
-a always,exclude
-a always,exclude
[william@t14 ~]$ sudo auditctl -R test-audit-2.rules ; echo $? ; sudo auditctl -l
0
-a never,exclude
-a always,exclude
-a always,exclude
```

stevegrubb:
It is up to the rules to decide what to do. auditctl simply carries out the command. If you want to reload rules, start with '-D' which deletes all rules. This is why 10-base-config.rules start by deleting previous rules. I think this issue can be closed.

stevegrubb:
I don't think there is anything to fix here. If there is, please feel free to reopen.

wpyoga:
I have re-read the documentation, and I see now what you mean. I've submitted PR #292 to improve the documentation.

Sorry for the delay in my response, and thank you for explaining the mechanism.

PR #292, opened 2023-01-15
Explain the -R option more precisely and give an example

wpyoga:
Following up the discussion at #279 , I think the documentation can be improved in order to remove any ambiguity. The file read in by the `-R` option contains commands instead of rules.

For example, in 10-base-config.rules, there is a line `-b 8192` -- it is a command, not a rule.

I've also added an example that explicitly mentions the need to execute the '-D' command in order to clear any preexisting rules.

The above is based on my understanding of how auditctl operates, please correct me if I'm wrong.

### DIFF
--- a/audisp/plugins/ids/ids_config.h
+++ b/audisp/plugins/ids/ids_config.h
@@ -37,7 +37,7 @@
 // sysctls, selinux booleans
 // update specific rpm, all rpms
 // restart service
-// drop service timed <- need to whitelist these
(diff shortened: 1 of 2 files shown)


wpyoga:
I synced the branch after making this pull request -- and now somehow it is closed. Maybe it's because I force-pushed the sync -- GitHub didn't explicitly say this, only that "you need to remove 2 commits".

I must have misunderstood the GitHub mechanism. Please see #293 for the reposted PR.

Issue #300, opened 2023-03-16
connect to the socket ”/var/run/audispd_events“, status is all  CONNECTING, no CONNECTED?


stillyger:
hi,  i find a problem.

The steps are as follows:
1. enable the af_unix plug-in；
2. run test process, it will connect the socket file ”/var/run/audispd_events“,  restart the test process at an interval of 5s；
3. after 1-2 hours，will see the problem, the connect status is all ”CONNECTING“, no ”CONNECTED“
```
[root@k8s-single audit]# netstat -anp | grep audi
unix  2      [ ACC ]     STREAM     LISTENING     390679   49098/auditd         /var/run/audispd_events
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  3      [ ]         STREAM     CONNECTED     390682   49098/auditd
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  2      [ ]         DGRAM                    390680   49098/auditd
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  2      [ ]         STREAM     CONNECTING    0        -                    /var/run/audispd_events
unix  3      [ ]         STREAM     CONNECTED     390681   49098/auditd
[root@k8s-single audit]#
```

/etc/audit/plugins.d/af_unix.conf:

```
active = yes
direction = out
path = builtin_af_unix
type = builtin
args = 0640 /var/run/audispd_events
format = string
```

/etc/audit/auditd.conf:

```
local_events = yes
write_logs = yes
log_file = /var/log/audit/audit.log
log_group = root
log_format = ENRICHED
flush = INCREMENTAL_ASYNC
freq = 50
max_log_file = 8
num_logs = 5
priority_boost = 4
name_format = NONE
##name = mydomain
max_log_file_action = ROTATE
space_left = 75
space_left_action = SYSLOG
verify_email = yes
action_mail_acct = root
admin_space_left = 50
admin_space_left_action = SUSPEND
disk_full_action = SUSPEND
disk_error_action = SUSPEND
use_libwrap = yes
##tcp_listen_port = 60
tcp_listen_queue = 5
tcp_max_per_addr = 1
##tcp_client_ports = 1024-65535
[...]

PR #301, opened 2023-03-21
add pthread mutex for fdchanges and fdchangecnt

stillyger:
to fix issue 300
[](https://github.com/linux-audit/audit-userspace/issues/300)

1.we try to add pthread mutex for fdchanges and fdchangecnt;
2.in fd_change we set fdchanges  first, then set ++fdchangecnt, so when fd_reify enter and get fdchangecnt, the fdchanges is the true value;

please check whether the modification scheme is appropriate？

### DIFF
--- a/src/libev/ev.c
+++ b/src/libev/ev.c
@@ -215,6 +215,7 @@
 #include <limits.h>
 
 #include <signal.h>
+#include <pthread.h>
 
 #ifdef EV_H
 # include EV_H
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for the patch. I'd rather leave libev pristine so that I do not have to apply lots of things when it gets upgraded. I'm sure the patch solves the problem. But I'd rather rework it so one thread talks to libev and there is some kind of messaging between threads to ask the main one to do something on behalf of the other one.

stillyger:
After thinking about it, i agree with you. It's true that modifying libev is not a good solution，and it will cause difficulties in subsequent update and maintenance. And Continue to study the right answer that only make calls to libev form main thread, it feels like a big modification.

stevegrubb:
This should be straight forward to fix. I'll try to find time over the weekend to push an update.

PR #303, opened 2023-04-03
Add context for UIDs and GIDs in audit event

dimv36:
I found a problem when parsing audit events via auparse in interpert mode.

Suppose we have an registered event in audit.log:

<pre>
type=ANOM_LOGIN_FAILURES msg=audit(1680165235.344:750): pid=16619 uid=0 auid=1000 ses=7 msg='pam_faillock uid=1003  exe="/usr/bin/login" hostname=arm1 addr=? terminal=pts/2 res=success'<0x1d>UID="root" AUID="administrator" UID="testuser-16293"
</pre>

Interpreted values for this event are registered well, but logic of parse events translates field `uid=1003 ` to `uid=root`, that is error. So, it's error due to cache logic in `static const char *lookup_uid(const char *field, uid_t uid)`. Logic of this function uses cache for returning already interpreted value, but can not used for events, that contains different values for one field (in our case `UID` - see row 212-214).

So, this patch forms interpreted keys as concatenation of user attribute (`UID` or `GID` and real value for it, e.g. `1000`, `1001`). So, auparse will use those keys for good replacement UID and GID fields via interpeted values.

### DIFF
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -3197,8 +3197,17 @@ char *auparse_do_interpretation(int type, const idata *id,
 
 	// Check the interpretations list first
 	if (interpretation_list_cnt()) {
+		int found = 0;
+
 		nvlist_first(&il);
-		if (nvlist_find_name(&il, id->name)) {
+		if (type == AUPARSE_TYPE_UID ||
+			type == AUPARSE_TYPE_GID) {
+			found = nvlist_find_name_ex(&il, id->name, id->val);
+		} else {
+			found = nvlist_find_name(&il, id->name);
+		}
+
+		if (found) {
 			nvnode* node = &il.array[il.cur];
 			const char *val = node->interp_val;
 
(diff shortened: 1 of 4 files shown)


stevegrubb:
Thanks for reporting this. It does look wrong. I'm thinking that pam should be using suid=1003. There should not be 2 uid fields in the same record. In the mean time, we probably need a workaround like your patch until pam ships a corrected update.

stevegrubb:
I filed https://github.com/linux-pam/linux-pam/pull/591 to fix this upstream.

dimv36:
Good job, Steeve. I think it will be great to have some developer's guide for using audit API in applications like PAM (so, what fields you can used, types, events and so on)

stevegrubb:
Thanks for reporting this. I filed bz against RHEL components and upstream pam has this fixed in the 1.6.0 release. I'm going to close this. The problem should work itself out without needing to patch the audit code.

PR #306, opened 2023-04-27
Check for AF_UNIX unnamed sockets

ematsumiya:
According to `man 7 unix`, an unnamed (or anonymous) AF_UNIX socket will have the following properties:

- its address length will be sizeof(sa_family_t)
- sun_family is AF_LOCAL/AF_UNIX (0x1) (obviously, but see below)
- sun_path must not be inspected (empirical evidence suggests sun_path[0] is 0 most of the time though)

Thus, the hex representation of such sockaddr_un is '0100'.  Which is observable as "saddr=0100" in SOCKADDR messages, or "len too short" when interpret flag '-i' is passed.

Example error messages:
```
# aureport -x --summary

Executable Summary Report
=================================
total  file
=================================
sun_path len too short
sun_path len too short
...
sun_path len too short
sun_path len too short
9426  /usr/bin/grep
7070  /usr/sbin/nmbd
2461  /usr/bin/rm
...
# ausearch -m SOCKADDR
...
type=SOCKADDR msg=audit(1681493602.806:85518): saddr=0100
...
# ausearch -m SOCKADDR -i
...
type=SOCKADDR msg=audit(04/14/2023 14:20:54.930:82007) : saddr={ saddr_fam=local sockaddr len too short }
...
```

This commit changes the interpreted message for unnamed/anonymous sockets, and make sure to not return an error if so.

Also, while at it, check if sun_path[1] is valid for abstract sockets cases before trying to strdup() it.

### DIFF
--- a/auparse/interpret.c
+++ b/auparse/interpret.c
@@ -1235,9 +1235,10 @@ static const char *print_sockaddr(const char *val)
         switch (saddr->sa_family) {
                 case AF_LOCAL:
 			if (slen < 4) {
-				rc = asprintf(&out,
-				    "{ saddr_fam=%s sockaddr len too short }",
-							str);
+				rc = asprintf(&out, "{ saddr_fam=%s %s }", str,
+				    slen == sizeof(saddr->sa_family) ?
+				    "unnamed socket" : // ignore sun_path
+				    "sockaddr len too short");
 				break;
 			} else {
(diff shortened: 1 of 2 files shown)


stevegrubb:
Thanks for the patch. In ausearch-parse.c, I almost feel like there should be something like (unnamed)  in the area of line 1735.

PR #310, opened 2023-06-15
Update ausearch.8 to include common example

lewisdenny:
This patch adds an example to ausearch man page for checking denials today, this is a common use for people visiting the man page.

### DIFF
--- a/docs/ausearch.8
+++ b/docs/ausearch.8
@@ -233,6 +233,11 @@ The boot time option is a convenience function and has limitations. The time it 
 
 date -d "`cut \-f1 \-d. /proc/uptime` seconds ago"
 
+.SH EXAMPLE
+.nf
+Check the SELinux log for any denials today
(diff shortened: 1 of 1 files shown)


lewisdenny:
@stevegrubb Hi mate, would you be able to review my patch if you get any free time? Thank you!

stevegrubb:
Hello...First, thanks for trying to improve the audit documentation. My hesitation on this PR is that it only gives an example for selinux AVCs. Some big distributions do not ship selinux. So the given example is not really applicable to the whole community. If there were other examples, it would be more useful to the community.

lewisdenny:
Thanks for your input Steve, that's a great point. Let me think about it some more as I think having some examples is important.

stevegrubb:
Planning to do a new release this weekend. If you can add a couple other examples to the pull request, I'll include it in the next release.

PR #311, opened 2023-06-25
Replace __attribute_malloc__ with __attribute__((__malloc__))

bkuhls:
__attribute_malloc__ is not available on musl

Fixes
| ../../git/auparse/auparse.h:54:2: error: expected function body after function declarator
|         __attribute_malloc__ __attr_dealloc (auparse_destroy, 1);
|         ^

Downloaded from https://cgit.openembedded.org/meta-openembedded/tree/meta-oe/recipes-security/audit/audit/0001-Replace-__attribute_malloc__-with-__attribute__-__ma.patch

### DIFF
--- a/auparse/auparse.h
+++ b/auparse/auparse.h
@@ -32,6 +32,9 @@
 # define __attr_dealloc(dealloc, argno)
 # define __attr_dealloc_free
 #endif
+#ifndef __attribute_malloc__
+#  define __attribute_malloc__
+#endif
 
 #ifdef __cplusplus
 extern "C" {


stevegrubb:
Thanks for the patch. But I think the right fix is to provide a missing define at line 31 for __attribute_malloc__. Should have been
``#ifndef __attribute_malloc__
    #define __attribute_malloc__
#endif
``

bkuhls:
> Thanks for the patch. But I think the right fix is to provide a missing define at line 31 for **attribute_malloc**. Should have been
> 
> ```
> #ifndef __attribute_malloc__
>     #define __attribute_malloc__
> #endif
> ```

Done

stevegrubb:
Thanks for the patch.

thesamesam:
> Thanks for the patch. But I think the right fix is to provide a missing define at line 31 for **attribute_malloc**. Should have been
> 
> ```
> #ifndef __attribute_malloc__
>     #define __attribute_malloc__
> #endif
> ```
> 
> The reason to not actually define it (so that it works) is that all of the attributes have a complicated version check to see if it should be supported. So, I'd rather just do a high level check so I don't have to add all that code and maintain it myself.

This isn't particularly clear to me. `__attribute__ ((__malloc__))` is what `__attribute_malloc__` is defined to be in glibc's headers anyway.

If you want to rely on glibc's own version checks, I could kind of understand that, but their check is:
```
/* At some point during the gcc 2.96 development the `malloc' attribute
   for functions was introduced.  We don't want to use it unconditionally
   (although this would be possible) since it generates warnings.  */
#if __GNUC_PREREQ (2,96) || __glibc_has_attribute (__malloc__)
# define __attribute_malloc__ __attribute__ ((__malloc__))
#else
# define __attribute_malloc__ /* Ignore */
#endif
```

i.e. it's for >=gcc-2.96. I suspect those compilers aren't tested anywhere and aren't actually needed for linux-audit.

PR #313, opened 2023-07-13
Fix the error found by clang-tidy

DmitryTD:
auditd-reconfig.c: In function 'start_config_manager': auditd-reconfig.c:63:42: warning: the comparison always evaluates to false because pthread_create always returns non-negative values


stevegrubb:
It would be so much easier if the man page said it returns errno. Thanks for the patch.

PR #316, opened 2023-08-01
audisp/af_unix.conf: Update default config values

travier:
For `path = builtin_af_unix`, fixes:
```
auditd[981]: Option builtin_af_unix line 10 is obsolete - using /sbin/audisp-af_unix
```
See: https://github.com/linux-audit/audit-userspace/blob/2cf5c2e0f72e5977b1fcbb136d084a041d06e4fd/audisp/audispd-pconfig.c#L397

For `type = builtin`, fixes:
```
auditd[981]: Option builtin line 11 is obsolete - update it
```
See: https://github.com/linux-audit/audit-userspace/blob/2cf5c2e0f72e5977b1fcbb136d084a041d06e4fd/audisp/audispd-pconfig.c#L432

### DIFF
--- a/audisp/plugins/af_unix/af_unix.conf
+++ b/audisp/plugins/af_unix/af_unix.conf
@@ -1,4 +1,3 @@
-
 # This file controls the configuration of the
 # af_unix socket plugin. It simply takes events
 # and writes them to a unix domain socket. This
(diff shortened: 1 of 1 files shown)


travier:
I've not tested that change yet. We could also consider completely removing those values as those are apparently default values.

stevegrubb:
Not sure how that got missed last release. Thanks!

PR #319, opened 2023-08-14
delete af_unix.conf in audit package

stillyger:
I found that the af_unix.conf does not need to exist in the main package. And the file is already in the audispd-plugins package.


stevegrubb:
I don't know how that keeps coming back. But yeah. Delete it.

PR #320, opened 2023-08-15
ausearch: do not use built-in logs when --input is specified

Cropi:
When processing a file using ```ausearch -if file_name``` as non-root user, the following is displayed:
```
$ ausearch -if ./test
Error opening config file (Permission denied)
NOTE - using built-in logs: /var/log/audit/audit.log
<no matches>
```

With patch applied:
```
Error opening config file (Permission denied)
NOTE - using built-in end_of_event_timeout: 2
<no matches>
```

The behavior is not changed, but correct information about configuration options are displayed to the user.
In case you are not able to view the config file, ```end_of_event_timeout``` can't be queried.

### DIFF
--- a/src/ausearch.c
+++ b/src/ausearch.c
@@ -109,9 +109,20 @@ int main(int argc, char *argv[])
 	(void) umask( umask( 077 ) | 027 );
 
 	/* Load config so we know where logs are and eoe_timeout */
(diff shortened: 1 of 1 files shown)


stevegrubb:
Thanks for the patch!

PR #327, opened 2023-10-12
Make session id consistently signed

tautschnig:
This fixes type-conflicting definitions and declarations.

Reported by CBMC's goto-cc compiler, which performs type-aware linking.

I should add that I am somewhat unsure about a change introduced in 1a0bfcc9a449: it parsed negative session ids with base `0`, which would suggest that inputs must be base 16 (but only in the negative case)?!

### DIFF
--- a/src/aureport-options.c
+++ b/src/aureport-options.c
@@ -61,7 +61,8 @@ const char *event_uuid = NULL;
 const char *event_vmname = NULL;
 long long event_exit = 0;
 int event_exit_is_set = 0;
-int event_ppid = -1, event_session_id = -2;
+pid_t event_ppid = -1;
+uint32_t event_session_id = -2;
 int event_debug = 0, event_machine = -1;
 time_t	arg_eoe_timeout = (time_t)0;
 
--- a/src/ausearch-options.c
+++ b/src/ausearch-options.c
@@ -895,19 +895,21 @@ int check_params(int count, char *vars[])
 			size_t len = strlen(optarg);
 			if (isdigit(optarg[0])) {
 				errno = 0;
-				event_session_id = strtoul(optarg,NULL,10);
-				if (errno)
+				unsigned long optval = strtoul(optarg,NULL,10);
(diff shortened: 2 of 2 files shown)


stevegrubb:
'0' is a special value that tells strtol to figure out what the base is. Usually it will be base 10, but people do strange things.

The kernel variables are uint32_t. If you do not use size based types, it can (and has) introduce problems on i386 systems. Since people have largely moved to x86_64 these days, the likelihood of introducing 32 bit problems can be higher than in the past since we have fewer testers. The code we have is working for 32 bit systems, so I'll leave this patch for another time when I can look it over more carefully. I just wanted to let you know 32 bit systems are the main concern in case it changes your patch.

tautschnig:
> '0' is a special value that tells strtol to figure out what the base is. Usually it will be base 10, but people do strange things.
> 

Ah, my bad for misparsing the man page.

> The kernel variables are uint32_t. If you do not use size based types, it can (and has) introduce problems on i386 systems. Since people have largely moved to x86_64 these days, the likelihood of introducing 32 bit problems can be higher than in the past since we have fewer testers. The code we have is working for 32 bit systems, so I'll leave this patch for another time when I can look it over more carefully. I just wanted to let you know 32 bit systems are the main concern in case it changes your patch.

Oh, more than happy to adjust my patch: I just went for the type that `src/aureport-options.c` uses, but could have gone either way. So we could consistently use `uint32_t` (as much as that's confusing with negative numbers), or `int`, or some other type (perhaps `int32_t`). But, as is, the types aren't consistent across `ausearch-*` and `aureport-options.c`.

stevegrubb:
Yes, I am aware that the typing is inconsistent. It has to be treated as uint32_t as much as possible, but there are times when you need it converted to treat it as a negative number.

tautschnig:
I have now updated the patch to consistently use `uint32_t`.

stevegrubb:
Thanks. Looks good.

PR #329, opened 2023-10-31
libev fixes

cgzones:


stevegrubb:
Thanks for the patch. For loop_init, we pass EVFLAG_NOENV which avoids the code being patched. (Another problem is it's using atoi which hides errors.) I try not to patch libev much because changes can get lost on the next libev upgrade.

About the io_uring backend, if that gets touched at all, I wonder if it would be better to simply not compile it? We only ever use select and epoll backends. So it would probably be best to modify the define of EV_USE_IOURING to 0 so all io_uring code disappears and auditd is around 27k smaller.

stevegrubb:
commit 3eb76aa should make part of this patch unnecessary.

stevegrubb:
My thinking is this should really go to upstream libev. The audit project avoids the problems you found. But if I fix it here, I have to remember to re-fix it next update to libev. If you do take it to libev upstream, he is somewhat problematic to work with. But it's worth a try. I'm on that mail list and I would chime in to support the patch.

cgzones:
The first commit was already fixed upstream via http://cvs.schmorp.de/libev/ev_iouring.c?r1=1.24&r2=1.25&.

stevegrubb:
OK, perfect. If a new release is ever made, I'll pick it up. The io_uring code has several problems. We're probably staying away from that for a while. Thanks for the whole patch set. I took a stab at more code hygiene cleanups in the src and auparse directories.

PR #338, opened 2023-10-31
lib: cast to unsigned char for character test functions

cgzones:
Passing a value not representable by unsigned char is undefined behavior.


stevegrubb:
Hmm. Just checked in glibc's ctype.h and isdigit is declared as taking an "int" argument.

stevegrubb:
ok, having looked into this more, I'll merge it. Thanks.

PR #339, opened 2023-10-31
Avoid file descriptor leaks in multi-threaded applications

cgzones:


stevegrubb:
This patch seems to have been filed out of order or written independently of the other pull requests. It has conflicts that need to be resolved. Do you want to rebase the patch on the current master branch and resubmit?

stevegrubb:
Thanks for fixing it up!

PR #342, opened 2023-11-01
Update syscall tables to match Linux 6.6 definitions

hillu:


stevegrubb:
OK. This looks good. But the 6.7 syscalls are already out, too (futex_*).

PR #346, opened 2024-01-08
ausearch: add option to exclude by message type

cgzones:
Similar to the existing option `-m`, which searches by the given message type(s), add the option `-M` to exclude by the given message type(s).


stevegrubb:
We could search by negating virtually any of the search criteria. I wonder if it would be more useful to add --negate and do it to everything? Is there a practical use to removing just one particular event type and selecting all others?

Also, -m supports lists.  ausearch -m DAEMON_START,DAEMON_END,SYSTEM_BOOT,SYSTEM_SHUTDOWN,SYSTEM_RUNLEVEL,SERVICE_START,SERVICE_STOP

Would the -M option do the same thing?

cgzones:
> Also, -m supports lists. ausearch -m DAEMON_START,DAEMON_END,SYSTEM_BOOT,SYSTEM_SHUTDOWN,SYSTEM_RUNLEVEL,SERVICE_START,SERVICE_STOP
> Would the -M option do the same thing?

Yes, since it reuses the same parsing logic.

> Is there a practical use to removing just one particular event type and selecting all others?

It seems useful while already filtering by a custom `key`, to reduce the shown events.

PR #347, opened 2024-01-15
lib: avoid UB on sequence wrap-around

cgzones:
Signed integer overflow is undefined, allowing compilers to optimize the condition `++sequence < 0` away.


stevegrubb:
Thanks for the patch

PR #348, opened 2024-01-15
lib: use thread-safe passwd and group lookup

cgzones:
Use getpwuid_r(3) and getgrnam_r(3) instead of their non thread-safe version to protect against concurrent usage, not necessarily within libaudit.

### DIFF
--- a/lib/libaudit.c
+++ b/lib/libaudit.c
@@ -1964,15 +1964,43 @@ void audit_rule_free_data(struct audit_rule_data *rule)
 	free(rule);
 }
 
+static void memset_secure(void *buf, size_t size)
+{
+	buf = memset(buf, '\0', size);
+	__asm__ __volatile__ ("" : : "r"(buf) : "memory");
+}
+
 // This use is OK because its creating rules for the local
 // machine and is looking up a local user.
 static int audit_name_to_uid(const char *name, uid_t *uid)
 {
-	struct passwd *pw;
+	struct passwd pwd;
+	struct passwd *result;
+	char *buf;
+	long bufsize;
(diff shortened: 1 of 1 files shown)


stevegrubb:
To my knowledge, these functions are only used by auditctl. Since the kernel only knows numbers, we convert names/groups in the rules to numbers. I'm wondering if I should mark a bunch of these functions deprecated and then pull them into a private library some time in the future?

cgzones:
There seem to be several third-party users of `audit_rule_fieldpair_data(3)`:
- https://github.com/ElementAI/osquery/blob/04b51fd450b699cc24f46604fc4576240b7a06e1/osquery/events/linux/audit.cpp#L271
- https://github.com/Azure/Azure-IoT-Security-Agent-C/blob/43bbc87ee8c1061925835d01f0b4cc1dbd089874/src/agent/src/os_utils/linux/audit/audit_control.c#L87
- https://github.com/buzz/e4rat/blob/7d2f33b578afc8df8ab7e3cd5ab77668aab53fc4/src/listener.cc#L186
- https://github.com/tricktux/file-audit-system/blob/4b302ca7b81b388b3826f3125b58c677e0f40db9/src/monitor.cpp#L62
- https://github.com/wazuh/wazuh/blob/0e4c42ae116e9d226a9000d359f1f0704175fc66/src/shared/audit_op.c#L314

stevegrubb:
Thanks for the info. I never would have guessed. :-)  I'll review this more tomorrow.

stevegrubb:
OK, looks good to me. Thanks!

PR #349, opened 2024-01-17
audispd: Variable amount of arguments

Cropi:
Every plugin configuration line comprises a key = value pair, optionally followed by an additional option at the line's end. Previously, if the value consisted of multiple words, an error would occur. This commit addresses the issue, allowing for the specification of multiple space-separated arguments in a plugin configuration. For instance, the audisp-syslog plugin can handle 3 arguments, but we were not able to pass all 3 at a single time due to this limitation.

```
args = LOG_NOTICE LOG_LOCAL0 interpret
```

### DIFF
--- a/audisp/audispd-pconfig.c
+++ b/audisp/audispd-pconfig.c
@@ -38,15 +38,15 @@
 struct nv_pair
 {
 	const char *name;
-	const char *value;
-	const char *option;
+	char **values;
+	int nvalues;
 };
 
 struct kw_pair
 {
 	const char *name;
 	int (*parser)(struct nv_pair *, int, plugin_conf_t *);
-	int max_options;
+	int max_options; /* -1 means any number of options */
 };
 
 struct nv_list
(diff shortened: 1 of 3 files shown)


stevegrubb:
Do we leak the memory from argv = calloc() before the execve? Otherwise looks good.

stevegrubb:
OK, looked at this more. I see there is a free on the parent's path out. That is good and alleviates that concern. Since this is a daemon, we need to minimize chances of heap fragmentation. I'd suggest moving the calloc() down after the fork in the child's code so as not to stir the parent's heap.

Cropi:
Sorry, I was on vacation at the end of last week. I also modified the code according to your suggestion.

> OK, looked at this more. I see there is a free on the parent's path out. That is good and alleviates that concern. Since this is a daemon, we need to minimize chances of heap fragmentation. I'd suggest moving the calloc() down after the fork in the child's code so as not to stir the parent's heap.

stevegrubb:
Thanks!

PR #350, opened 2024-01-24
Make the paths inside the systemd .service files configurable

bigon:


stevegrubb:
I think that instead of changing Makefile.am, all you need to do is add 
init.d/auditd.service init.d/audit-rules.service 
into AC_CONFIG_FILES() in configure.ac. Autoconf should be able to take it from there.

bigon:
@stevegrubb I tried to do that, and it's not working, it's replacing `@sbindir@` by `${exec_prefix}/sbin`

It's documented here: https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Installation-Directory-Variables.html


> Similarly, you should not rely on AC_CONFIG_FILES to replace bindir and friends in your shell scripts and other files; instead, let make manage their replacement. [...]

So using sed in make is the recommended way

stevegrubb:
I see that, too, if I stop the build or by hand only do the compile. If it continues all the way to install, it turns out with the final substitutions in place. When I install the resulting package, it has /usr/sbin/auditd instead of ${exec_prefix}/sbin/

bigon:
@stevegrubb It's definitely not working here.

`ExecStart=@sbindir@/auditd` in the `init.d/auditd.service.in` is translated to `ExecStart=${exec_prefix}/sbin/auditd` after the configure.

And the value is the same after I run `DESTDIR=/tmp/foo make install`, so `${exec_prefix}` is never expanded for me

stevegrubb:
Actually, we do need to touch Makefile.am. It needs:
DISTCLEANFILES = auditd.service audit-rules.service

Rechecking things, if I do a make dist, copy that tar file somewhere and unpack it, cd into the dir, run configure, make, make check, make install, it continues having the ${prefix}. If instead I use the tar file with rpm to do a build, init.d/audit.service has correct substitutions. The only difference is I didn't add --sbindir, --prefix, --localstatedir first time. When I add them to the configure command, the substitutions are correct before ever doing make.

Fedora is using autoconf-2.71 and automake-1.16.5. The notes for 2.71 still gives the same advice to use sed which was pointed out in the autoconf manual page previously. Hmm...

bigon:
Yes I think it makes sense if you explicitly set the --*dir flags, but if you keep the default, the ´${exec_prefix}´ is not expanded

stevegrubb:
In general, I like the idea behind this patch. I guess we can't rely on defaults getting substituted. I'll try to find time in the next day or two to test this with "make distcheck".

stevegrubb:
OK, I have some time to look at this again. If I add the patch and run autogen.sh, it complains:
init.d/Makefile.am:41: warning: '%'-style pattern rules are a GNU make extension

If we ignore that and do make distcheck it ends like this:
/usr/bin/install -c -D -m 644 ../../../init.d/auditd.service /home/sgrubb/audit/audit-4.0.1/_inst/lib/systemd/system
/usr/bin/install: cannot stat '../../../init.d/auditd.service': No such file or directory
make[5]: *** [Makefile:631: install-exec-hook] Error 1

It has to pass the make distcheck test to be applied. This is how I always make the release to be sure that everything is correctly done.

bigon:
@stevegrubb can you please retry?

I fixed the installation of the file

stevegrubb:
Thanks! This works much better. Merging.

PR #352, opened 2024-02-14
audisp-filter: new plugin that provides audit event filtering capability

Cropi:
### Description
Audit plugins are components used in the Linux audit system to extend its functionality and customize the way audit events are processed and handled on userspace level. One common feature that all plugins lack is the ability to filter out specific events. The goal of this PR is to introduce a new plugin, capable of filtering out specific audit events based on the configuration passed to. Furthermore, it's able to forward remaining logs to the configured plugin, operating in both allowlisting and blocklisting modes.

### Implementation details
The `audisp-filter` program is accompanied by an associated config file containing `ausearch-expression(5)` s. Each line represents an  expression that can be either a simple rule or a very complex expression. These rules are grouped together using the OR operator to form one complex expression.

Once the rules are successfully loaded into an inter list of expressions, a fork happens. The parent process receives unfiltered audit events, directing them to the filtering engine. Here, audit records are searched using the expression created at the beginning. The decision to forward an audit event to the configured plugin depends on two factors: the operational mode (allowlist, blocklist) of audisp-filter and whether the expression matches the ongoing event.
On the other hand, the child process will execute the configured binary (e.g. `audisp-syslog`) and receive filtered audit events on its standard input.

Example audisp plugin configuration:
```
active = yes
direction = out
path = /sbin/audisp-filter
type = always 
args = allowlist /etc/audit/filter.conf /sbin/audisp-syslog LOG_USER LOG_INFO interpret
format = string
```

Example plugin specific configuration:
```
# Skip the openat syscall
(type r= SYSCALL && syscall i= openat)
```

I've attached a man page as well. Any  feedback will be much appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ cscope.*
 audisp/audispd
 audisp/plugins/af_unix/audisp-af_unix
 audisp/plugins/remote/audisp-remote
+audisp/plugins/filter/audisp-filter
 audisp/plugins/syslog/audisp-syslog
 audisp/plugins/zos-remote/audispd-zos-remote
 audisp/plugins/statsd/audisp-statsd
--- a/audisp/plugins/Makefile.am
+++ b/audisp/plugins/Makefile.am
@@ -23,7 +23,7 @@
 
 CONFIG_CLEAN_FILES = *.loT *.rej *.orig
 
-SUBDIRS = af_unix remote syslog
+SUBDIRS = af_unix remote syslog filter
 if ENABLE_EXPERIMENTAL
 SUBDIRS += ids statsd
 endif
--- a/audisp/plugins/filter/Makefile.am
+++ b/audisp/plugins/filter/Makefile.am
@@ -22,11 +22,12 @@
 #
 
 CONFIG_CLEAN_FILES = *.loT *.rej *.orig
-EXTRA_DIST = filter-syslog.conf $(man_MANS)
+EXTRA_DIST = filter.conf audisp-filter.conf $(man_MANS)
 AM_CPPFLAGS = -I${top_srcdir} -I${top_srcdir}/lib -I${top_srcdir}/common -I${top_srcdir}/auparse
 prog_confdir = $(sysconfdir)/audit
+prog_conf = audisp-filter.conf
 plugin_confdir=$(prog_confdir)/plugins.d
-plugin_conf = filter-syslog.conf
+plugin_conf = filter.conf
 sbin_PROGRAMS = audisp-filter
 man_MANS = audisp-filter.8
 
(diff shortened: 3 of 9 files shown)


stevegrubb:
Hello,  I was reading through the patch. In handle_events, why do you malloc the event out of the passed auparse state and then create a new instance of auparse state taking the new buffer? The event in the passed state should be searchable. If it can't be searched, that should be looked into as a separate PR that this one depends on.

stevegrubb:
Should the filter process also relay signals to the child? For example, maybe the HUP signal. Also, if the execve fails due to a typo or selinux AVC, should it notify the parent? Otherwise the parent is handling events for no purpose. Also, I think it needs a sigchld handler which calls waitpid to reap child processes.

Cropi:
> Hello, I was reading through the patch. In handle_events, why do you malloc the event out of the passed auparse state and then create a new instance of auparse state taking the new buffer? The event in the passed state should be searchable. If it can't be searched, that should be looked into as a separate PR that this one depends on.

The auparse state is searchable. I duplicated it because, when you use `ausearch_next_event` on the current audit event and the event doesn't match the expressions provided to auparse state, there seems to be no direct method (that I'm aware of) to navigate to the beginning of the audit event list. This is needed in order to be able to forward the event after searching the event.

stevegrubb:
auparse_first_record should take you back to the beginning since there is only 1 event to look at using auparse_feed.

Cropi:
I updated the PR to contain `ausearch_cur_event` commit from the other branch to see it in action. Will remove it once it's merged.

stevegrubb:
ausearch_cur_event patch was merged. This may need an update.

Cropi:
> OK, I think we are getting close. I would recommend moving the setup of the search to right after the auparse_add_callback(). It should only need to be setup once and then used in the handle_event function. The ausearch_clear should be done near the auparse_destroy.
> 
> In the child handler, it might be better to just have an empty loop. IOW, just ';'

I accidentally left there that debug printf :0

> 
> Compile-wise, we have these: audisp-filter.c:150:33: warning: too many arguments for format [-Wformat-extra-args] needs %s audisp-filter.c:255:13: warning: unused variable ‘rc’ [-Wunused-variable] delete audisp-filter.c:217:13: warning: ‘print_rules’ defined but not used [-Wunused-function] Comment out or delete
> 
> As a project, we try to follow the Linux Kernel programming style for consistency. With the above fixups, I think we can merge the patch.

I've updated the patch to contain you suggestions. Thanks for the feedback.

stevegrubb:
Just did a build with the patch applied. It complained that filter-syslog.conf is unpackaged. However, everything else is named audisp-filter. I think the conf file needs renaming before merging. The man page mentions filter.conf which is also fine. The syslog-filter.conf file references /sbin/audisp-syslog. Also, isn't there some argument that is supposed to point to a rules file?

Please double check these items. I think we're OK on the code, just need to tidy up file names in the configuration.

Cropi:
Okay, I did some cleanup. Just for reference I will leave a comment here. There are two configs:

- **General plugin config** - `/etc/audit/plugins.d/filter.conf` - The args parameter must include a path to the plugin-specific config.
- **Plugin specific config** - `/etc/audit/audisp-filter.conf` - will contain list of ausearch-expressions that are gonna be added to the filter engine

stevegrubb:
Thanks for all the work! It's ready to merge.

PR #357, opened 2024-02-22
libaudit: add ausearch_cur_event

Cropi:
The `ausearch_cur_event` function will scan the input source and evaluate whether any record in the current event contains the data being searched for.

The first part of `ausearch_cur_event` and `ausearch_next_event` are identical, so it can be moved to a separate function to remove duplicate code if needed.

This will help us to progress with https://github.com/linux-audit/audit-userspace/pull/352

### DIFF
--- a/auparse/auparse.c
+++ b/auparse/auparse.c
@@ -1439,7 +1439,7 @@ int ausearch_cur_event(auparse_state_t* au) {
 		} else if (rc < 0)
 			return rc;
 	}
-	ausearch_reposition_cursors(au);
+
 	return 0;
 }
 
--- a/auparse/auparse.c
+++ b/auparse/auparse.c
@@ -1420,6 +1420,30 @@ static int ausearch_compare(auparse_state_t *au)
 }
 
 // Returns < 0 on error, 0 no data, > 0 success
+int ausearch_cur_event(auparse_state_t* au) {
+	int rc, records;
+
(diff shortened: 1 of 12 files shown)


stevegrubb:
Thanks for the new function. In looking at the code and how ausearch_next_event is written, I think that you do not need to alter the started member of the expression struct. It appears it's only use is to call auparse_next_event in ausearch_next_event. Otherwise it's never used. Since this is knowingly dealing with only 1 event, I think it should do something like this:
`
int i, records = auparse_get_num_records(au);
for (i=0; i < records; i++) {
    auparse_goto_record_num(au, i);
    if ((rc = ausearch_compare(au)) > 0) {
        ausearch_reposition_cursors(au);
        return 1;
    } else if (rc < 0)
        return rc;
}
`
Of course this needs to handle any error from goto which I omitted for clarity. It should also return something appropriate if it finds nothing after all records are checked. (Also, there is a warning on the man page for auparse_next_record and feed callback functions. While the goto is less efficient, at least it won't deadlock.)

stevegrubb:
On success, yes we should reposition the cursors since the developer has a choice on where to stop. On no match, I'd document that a failed search will leave it positioned on the last record of the current event. The app developer can then decided if they need to go back to the first record.

stevegrubb:
OK. Looks good. Thanks!

PR #366, opened 2024-04-15
avoiding of NULL pointers dereference

Yugend:
I recently sent a similar PR, but missed two more potential dereferences of null pointers. 
In general, there are many points in the parse_user() function where dynamic memory allocation occurs. But the subsequent pointer dereference occurs only in one place. I do not know how critical this is, so I did not add many checks, I made only one obvious one.


stevegrubb:
Looks good. Thanks!

PR #371, opened 2024-05-03
normalize-llist.c: fix compile error

zpc0:
On Fedora 40 system, `-Wimplictit-function-declaration` error
is occur when compiling `normalize-llist.c` `printf()`.
Including `stdio.h` will fix this.

This can be applied to 3.1 branch.

### DIFF
--- a/auparse/normalize-llist.c
+++ b/auparse/normalize-llist.c
@@ -21,6 +21,7 @@
  *   Steve Grubb <sgrubb@redhat.com>
  */
 
+#include <stdio.h>
 #include <stdlib.h>
 #include "normalize-llist.h"
 


stevegrubb:
Thanks for the patch. This came from a recent PR. Auparse is a library and should not be calling printf. I have a feeling we should just delete the call to printf. Maybe this is a good reason to change cllist_append from void to int so its caller can return an error and that is how this failure gets communicated rather than printf.

stevegrubb:
Actually, there was quite a number of these to fix. I fixed it by returning an error code. Thanks for letting me know about this.

zpc0:
@stevegrubb Thank you for creating more better fix!
Since projects using release 3.1.4 such as OpenWrt are also affected by this issue, could you backport the fix to 3.1 branch and making the new release? Thanks!

PR #374, opened 2024-05-21
Remove systemd as runtime dependecy

kl4zo:
Auditd does not depend on `systemd` for building and running. It only depends on `systemd` if systemd service files are used to start `auditd`. But `auditd` can be started by other init systems e.g. OpenRC as well. As done in Alpine Linux (https://git.alpinelinux.org/aports/tree/main/audit/auditd.initd)

Therefore stating `systemd` as runtime dependency is misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ RUNTIME DEPENDENCIES
 * kernel >= 5.0 
 * systemd
 
+NOTE: While this repository provides support for systemd to start the audit
+daemon, other init systems can be used as well. For example, [Alpine
+Linux](https://git.alpinelinux.org/aports/tree/main/audit/auditd.initd) provides
+an init script for OpenRC.
+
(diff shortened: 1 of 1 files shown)


stevegrubb:
I've delayed responding to this to gather my thoughts. I've had my arguments with systemd developers here and in their issue tracker. At this point, I'm tired of it. I cannot accept this patch in it's current form because this repo literally provides systemd support. Not that I am an enthusiast, but it's reality.

What I am willing to accept is a statement that says something to the effect that auditd can be run by other init systems with a link to supporting documents and a link to an example file used by said init system that users of auditd would find useful. I want to find a middle ground that I publish useful information to auditd users without getting bogged down in systemd's merits.

I hope you understand.

kl4zo:
That sounds reasonable. I implemented the changes. LMK what you think.

PR #375, opened 2024-06-06
Fix CPPCHECK_WARNING findings in auparse/ellist and ausearch-parse

Cropi:
The `cppcheck` static code analysis tool reports two issues found in ausearch/auparse. Both of them seem to be false positives but it would be better to not see them.


stevegrubb:
Thanks for the patch. I understand this to be false positives. Closing.

PR #376, opened 2024-06-06
Fix USE_AFTER_FREE SAST findings

Cropi:
Covscan reports there are two issues labeled with USE_AFTER_FREE.
1. lru: In `remove_node`, when the queue has only one single element,. the `queue->front` pointer is updated but the `queue->end` is not. It does not cause any problems because in `dequeue` we loop until `queue->count > 0` , instead of `queue->end != NULL` .
2. auditd-event: covscan says that `Equality between "oconf->plugin_dir" and "nconf->plugin_dir" implies that they are aliases.` I think the condition `if (oconf->plugin_dir != nconf->plugin_dir` ...) will always be true because in `clear_config`, we have:
`config->plugin_dir = strdup("/etc/audit/plugins.d");`


Cropi:
This PR was just to silence covscan. Detailed explanation on the false positive: https://github.com/linux-audit/audit-userspace/commit/4939b8541322cbf3a53affc28e71ce53d92f121f 
Closing.

PR #379, opened 2024-06-10
Translate operator to symbol when displaying rules with exe/path/dir

Cropi:
### Problem description
1. Add a rule that contains `-F exe!=/some/path/`
2. List all rules using `auditctl -l`
3. The != operator is not gonna be correctly displayed.

This works fine internally, meaning the rule is correctly added to the kernel.


stevegrubb:
Looks good. Thanks!

Issue #382, opened 2024-06-29
Cannot install python swig module without python symlink installed


bigon:
Hello,

On debian the python -> python3 symlink is not available by default, and the installation of the swig module is failing with the following error:

```
../../../../../py-compile: line 125: test: found: integer expression expected
../../../../../py-compile: line 137: python: command not found
make[6]: *** [Makefile:523: install-pyexecPYTHON] Error 127
```

It seems that the PYTHON environment variable is not passed to the py-compile script

bigon:
Yes yes I did, it works.

But it only works because of the fallback in the py-compile script.

I proposed a MR (https://github.com/linux-audit/audit-userspace/pull/383) that fixes the root cause

PR #383, opened 2024-06-29
Do not try to override the value of the PYTHON variable with non-exis…

bigon:
…ting PYTHON3 one

Fixes: #382


Blarse:
Hi,

May be set PYTHON to ${PYTHON3:-$PYTHON} to override in case it's defined.

stevegrubb:
Thanks for the patch.

PR #385, opened 2024-07-08
Custom strndupa: Convert it to a macro instead of a function

### LINKED ISSUE #381: In-house implementation of `strndupa` is faulty
```
#ifndef HAVE_STRNDUPA
static inline char *strndupa(const char *old, size_t n)
{
	size_t len = strnlen(old, n);
	char *tmp = alloca(len + 1);
	tmp[len] = 0;
	return memcpy(tmp, old, len);
}
#endif
```

This code is undefined behavior, a pointer to stack memory is returned but the stack memory is reclaimed immediately. See our build for Void Linux https://github.com/void-linux/void-packages/pull/50666

Cropi:
Looking at
https://www.gnu.org/software/libc/manual/html_node/Truncating-Strings.html , `strndupa` is implemented as a macro. It uses `malloca`, which allocates space in the stack frame of the caller. The temporary space is automatically freed when the function that called alloca returns.

Fixes #381

### DIFF
--- a/auparse/auparse.c
+++ b/auparse/auparse.c
@@ -1204,14 +1204,13 @@ static int str2event(char *s, au_event_t *e)
 
 #ifndef HAVE_STRNDUPA
 #define strndupa(s, n)								\
-	(												\
 	({												\
 		const char *__old = (s);					\
 		size_t __len = strnlen (__old, (n));		\
 		char *__new = (char *) alloca(__len + 1);	\
 		__new[__len] = '\0';						\
 		(char *) memcpy (__new, __old, __len);		\
-	}))
+	})
 #endif
 
 /* Returns 0 on success and 1 on error */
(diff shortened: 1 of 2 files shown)


MIvanchev:
I assume this is from GLIBC because it looks quite like https://github.com/bminor/glibc/blob/master/include/string.h#L72. You don't need the outer `()` because GLIBC's version is using them to delimit the __extension__ keyword. Also keep in mind the construct `({....})` is also a GNU C extension: https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html. In short, it might be better to refactor the code so `strndupa` isn't a requirement any longer. This seems to be related to the function `audit_strsplit` (https://github.com/linux-audit/audit-userspace/blob/master/common/strsplit.c#L58) which is sadly stateful and comes with some problems itself.

stevegrubb:
This is almost a 20 year old project. Back in the old days, memory management wasn't as good and you could wind up with fragmented heaps. The hot path of auditd used as little memory allocation as possible or same sized memory allocation to avoid fragmented memory. At this point, due to record enrichment, we probably do a fair amount of allocating in handling each record.

stevegrubb:
I think we can drop the extra () and then I can merge it. I'll cherry pick this to the audit-3.1 branch.

Cropi:
I adjusted the patch by removing the extra `()`.

PR #393, opened 2024-09-05
audit.spec: remove unsupported '--enable-systemd'

ematsumiya:
Fixes: 7169631bef7c ("Update the spec file to be more correct")


stevegrubb:
Please see https://github.com/linux-audit/audit-userspace/pull/374

ematsumiya:
Hm AFAIU this  PR seems complementary to that?

Furthermore '--enable-systemd' is not defined anywhere, so in e.g. both openSUSE and Fedora I get:

> ./configure ... --enable-systemd
> ...
> configure: WARNING: unrecognized options: --with-python, --enable-systemd

My intention here was more of a cleanup than some feature addition/removal.

stevegrubb:
I see. Thanks.

PR #394, opened 2024-09-09
ausearch-parse: fix parsing for success/uid in parse_daemon1()

sergio-correia:
In `parse_daemon1()`, we may have the uid= field appear both before and after pid=, which may cause our parsing of it to fail, as we may have skipped past it. For uid=, let us search from the beginning.

Example for this case:

`type=DAEMON_END msg=audit(1709723032.140:753): op=terminate auid=0 uid=0 ses=8 pid=107086 subj=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 res=success`

`ausearch -if sample.log -a 753 -m DAEMON_END  -ui 0 --session 8 -p 107086 --success yes`

### DIFF
--- a/src/ausearch-parse.c
+++ b/src/ausearch-parse.c
@@ -1549,7 +1549,9 @@ static int parse_daemon1(const lnode *n, search_items *s)
 
 	// uid - optional
 	if (event_uid != -1) {
-		ptr = term;
+		// As the uid= field may happen in different orders, e.g. both before
+		// and after pid=, let us search for the uid from the beginning.
+		term = mptr;
 		str = strstr(term, " uid=");
 		if (str) {
 			ptr = str + 5;


stevegrubb:
Just a FYI, there is an ausearch test here: https://people.redhat.com/sgrubb/audit/ausearch-test-0.6.tar.gz

sergio-correia:
> Just a FYI, there is an ausearch test here: https://people.redhat.com/sgrubb/audit/ausearch-test-0.6.tar.gz

Thanks, I was not aware of this, although after a closer look, it seems it was a variation of that `ausearch-test` that identified the issues with the records reported here.

stevegrubb:
The daemon part checks OK (although I wonder if it best to just fix the event and let it work out over time).

Was the other fix only because of USER_MAC_POLICY_LOAD or are there other cases? The reason I ask is that there are 2 types of events that are fundamentally broken: USER_MAC_POLICY_LOAD and USER_AVC. Historically, they have been broken because libselinux doesn't create valid audit records.

The real fix would be to patch this: https://github.com/bus1/dbus-broker/blob/4fa73fd2dc8429ff0fbbda39bce8e8af92681b07/src/util/audit.c#L33  so that it does not use the avc logging function except when it really is an AVC. The other two cases should use audit_log_user_message(). If this is fixed at the source, is this patch still needed for USER_MAC_POLICY_LOAD?

sergio-correia:
Right, the other fix was due to `USER_MAC_POLICY_LOAD`, and I agree with you that the real fix would be in `dbus-broker`. Shall I remove that fix from this PR?

What about "just fix the event and let it work out over time"? Would you elaborate on that, please?

stevegrubb:
"just fix the event and let it work out over time" means that sometimes I have chosen not to fix something in current code and instead fix the problem in the event. Eventually the fixed event is in all distributions and then search works again. The idea is not to make ausearch slowed down by events that should be fixed. For example, maybe DAEMON_START should be re-arranged so that it fits all other DAEMON events and then it everything starts working next release (which should be soon).

A patch should be sent to dbus-broker. Around line 123 it should do something like 
if (type == UTIL_AUDIT_TYPE_AVC)
    audit_log_user_avc_message(...
else
    audit_log_user_message(...

Btw, I made a repo for ausearch-test here: https://github.com/stevegrubb/ausearch-test

sergio-correia:
Patch updated to have only the `parse_daemon1()` bits.

Cropi (approved):

PR #397, opened 2024-09-10
Change dependency from Requires to Wants for audit-rules.service

Cropi:
Replaced `Requires=audit-rules.service` with `Wants=audit-rules.service` in auditd.service to align with behavior from audit v4.0 and earlier. This modification ensures that if `audit-rules.service` fails to load rules into the kernel, `auditd` will not be forced to exit. This change preserves backward compatibility and prevents potential disruptions caused by issues in rule loading.


stevegrubb:
Thanks

PR #398, opened 2024-09-20
Skip plugin configs that do not have .conf suffix

Cropi:
While the naming format for plugin configuration files is documented in the auditd-plugins(5) man page, it may appear counterintuitive and deviates from the common naming conventions followed by other widely-used components (e.g., systemd, httpd, SELinux, dracut, and others).
This patch also addresses the handling of non-regular files.

With the patch applied, plugin configurations matching the following pattern would be skipped:

- not a regular file
- hidden file
- backup file
- file without .conf suffix


stevegrubb:
Thanks

PR #401, opened 2024-10-01
Add entry for uretprobe syscall (6.11, x86_&4-only)

hillu:


Cropi:
It seems like this is not a traditional syscall that should be directly invoked. Though interpretation could be added, otherwise the translated output will look like this:
```
At 04:24:02 10/02/2024 root unsuccessfully triggered-audit-rule uretprob-syscall-key using /root/test/main
```
I am not able to come up with a proper normalization name.

stevegrubb:
Is there a man page for this syscall? Enabling the syscall number may not be sufficient. We may need to add interpretations depending on the syscall.

PR #402, opened 2024-10-09
syscall list update: Add multiple missing syscalls

Cropi:
This PR deals only with syscall additions, which have not been added to the tables.
Originally part of https://github.com/linux-audit/audit-userspace/pull/389, but that will contain the renamings.


stevegrubb:
Thanks

PR #407, opened 2024-10-13
Correct printf format specifier as per issue 406

burnalting:
Need to set the format specifier of %lld for some prints of an event's seconds value as time_t is now int64_t. In case of back porting to a 32 bit time_t, the output is caste to (long long int) when printing.


Cropi (approved):
Based on https://github.com/linux-audit/audit-userspace/commit/14ccfbd5b05bb429defd7b5c6170995f8f23cb8e, I think it's safe to merge.

PR #411, opened 2024-10-18
ausearch format: Fix display of renamed file

Cropi:
In some cases, ausearch was not correctly showing the new name of a renamed file when searching for audit events. If the target file didn’t exist prior to the rename, ausearch was unable to parse the new file name. This occurred because  ausearch attempted to retrieve this information from the 7th record, which is absent when the target file does not exist.

Before patch:
```
At 08:37:22 10/18/2024 alakatos successfully renamed /home/alakatos/github/audit-userspace/test/r-r using /home/alakatos/github/audit-userspace/test/main
```

After patch:
```
At 08:37:22 10/18/2024 alakatos successfully renamed /home/alakatos/github/audit-userspace/test/r-r to /home/alakatos/github/audit-userspace/test/r-r.renamed using /home/alakatos/github/audit-userspace/test/main
```
The same goes for the csv output.

This will not fully fix the `rename` sycall "formatting problem". If the syscall is not successful, the output will still not be correct. But that will be part of another patch.

### DIFF
--- a/auparse/normalize.c
+++ b/auparse/normalize.c
@@ -693,7 +693,20 @@ static int normalize_syscall(auparse_state_t *au, const char *syscall)
 		case NORM_FILE_RENAME:
 			act = "renamed";
 			D.thing.what = NORM_WHAT_FILE; // this gets overridden
-			set_prime_object2(au, "name", 4);
+			/* A sucessfull syscall from the rename family will provide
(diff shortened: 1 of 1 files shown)


stevegrubb:
This fixes --format text. (csv is still missing.) Merging. Thanks!

PR #413, opened 2024-10-29
Log runlevel changes to console (during boot)

Cropi:
Audit may initiate runlevel changes in several scenarios, such as when available disk space for audit logs becomes critically low. In these situations, audit can halt the system during boot, but without clear evidence indicating that audit was responsible for the shutdown. With the patch applied, audit will write to `/dev/console` when doing runlevel changes.

This comes in handy during boot, so sysadmins can track down the root cause of the shutdown much faster.

### DIFF
--- a/audisp/Makefile.am
+++ b/audisp/Makefile.am
@@ -23,7 +23,7 @@
 
 SUBDIRS = plugins 
 CONFIG_CLEAN_FILES = *.rej *.orig
-AM_CPPFLAGS = -D_GNU_SOURCE -fPIC -DPIC -I${top_srcdir} -I${top_srcdir}/lib -I${top_srcdir}/src -I${top_srcdir}/src/libev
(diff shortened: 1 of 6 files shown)


stevegrubb:
The printf function attribute should be added to the function prototype. Look at audit_msg for an example. Also, normally we send messages after we perform the action in case the change fails. (There were cases in the past where something with init changed and it was unable to change to single user mode.) For halt, you have to send it before.

Cropi:
I put some comments under https://github.com/linux-audit/audit-userspace/pull/413/files#diff-561528b72c0692af52f6b4c90aae3f9ee3732c5837f448cddd68871a91cf6649R191. I was playing with options such as `space_left_action=` `SINGLE` and `HALT`. In case of SINGLE, I ended up in a deadlock. It's risky to call `check_space_left` during boot when auditd is not yet fully initialized.

PR #419, opened 2024-12-10
Add audit-tmpfiles.conf to ensure /var/lib/audit exists

cgwalters:
On image based systems for many years the idea has been to have the operating system content live in `/usr` as much as possible, and synthesize data outside there via external tools. For `/var` content, systemd-tmpfiles is a go to choice for situations like this.

For more information see
xref https://gitlab.com/fedora/bootc/tracker/-/issues/50

https://issues.redhat.com/browse/RHEL-70633

### DIFF
--- a/init.d/Makefile.am
+++ b/init.d/Makefile.am
@@ -23,7 +23,8 @@
 
 CONFIG_CLEAN_FILES = *.rej *.orig
 CLEANFILES = $(BUILT_SOURCES)
-EXTRA_DIST = auditd.service.in audit-rules.service.in auditd.conf auditd.cron \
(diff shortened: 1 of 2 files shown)


Cropi:
Hello, just a few nitpicks:

- In the description you refer to /var/lib/audit but you actually meant /var/log/audit, right?
- In Makefile.am, please add the new file to `EXTRA_DIST` so it gets included in the tarball as well
- In Fedora, the spec file is different from upstream (it's created from scratch), so that will still need to be changed.

cgwalters:
Thanks, addressed those points.

Cropi (approved):
LGTM!

PR #420, opened 2025-01-03
Change several log levels from NOTICE to INFO

andrew-sayers:
Lowering the priority of normal progress messages makes it easier for users to notice the messages they might need to do something about.

### DIFF
--- a/src/auditd-event.c
+++ b/src/auditd-event.c
@@ -755,13 +755,13 @@ static void check_log_file_size(void)
 				break;
 			case SZ_ROTATE:
 				if (config->num_logs > 1) {
-					audit_msg(LOG_NOTICE,
+					audit_msg(LOG_INFO,
 					    "Audit daemon rotating log files");
 					rotate_logs(0, 0);
 				}
 				break;
 			case SZ_KEEP_LOGS:
-				audit_msg(LOG_NOTICE,
+				audit_msg(LOG_INFO,
 			    "Audit daemon rotating log files with keep option");
 					shift_logs();
 				break;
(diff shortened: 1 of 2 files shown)


Cropi:
Hello, Why lower the priority of these messages? Are the logs getting too noisy with unnecessary entries?
This shouldn’t really happen since these messages are only sent to syslog during startup, reconfiguration, or when the logs are rotated.

andrew-sayers:
Er, yes that's certainly an important problem - the logs getting too noisy with unnecessary entries, because messages are sent to syslog on every startup, reconfiguration, and every time the logs are rotated.  Most importantly, when I reboot my computer and all my services want my attention at the same time, auditd progress messages crowd out other services' warning messages.

Another problem is that normal progress messages like "Init complete" are displayed the same as potentially significant problems like "Cannot open out of memory adjuster", making it harder to maintain the service.  Lowering the priority is particularly useful on systemd systems, because `sudo journalctl -p 0..5 -ru <service-name>` is usually enough to list just the actionable events, but auditd shows a page of normal progress messages with actionable events sprinkled in.

stevegrubb:
In the patch, the only normal progress message is that it's listening for events. No plugins found could signify a misconfiguration. I'd like to keep that. The log rotation messages are traditionally something people wanted to know. I suppose those could be lowered.

We're close to doing a release. If you drop the no plugins found, we can merge it for the next release.

andrew-sayers:
Fair enough, patch updated.  If "no plugins found" is still bothering me in a few months, I'll come back and bother you about adding a flag or something :)

stevegrubb:
Thanks. Most people use a plugin of some kind to get events to a central collector.

PR #422, opened 2025-01-06
Move $(datadir)/audit-rules into $(libdir)/audit-rules

Cropi:
The Linux Filesystem Hierarchy Standard (FHS) specifies that files under /usr/share/ should be architecture-independent. However, some distributions remove certain sample rules that are not applicable to specific architectures. As a result, the content of audit rules shipped may vary between architectures.

This patch ensures that $(datadir)/audit-rules will continue to exist as a symlink pointing to the new architecture-specific path.


Cropi:
Let's not complicate things by introducing symlinks to satisfy FHS. I am closing this.

PR #426, opened 2025-01-13
permtab fix: filter unsupported syscalls based on architecture

Cropi:
Filter syscalls to ensure compatibility with architecture-specific availability

Added logic to filter syscalls during rule creation to exclude unsupported syscalls based on the target architecture. Falls back to the original syscall list if filtering fails.

Reproducer on aarch64:
```
30-stig.rules:-a always,exit -F arch=b64 -F path=/etc/group -F perm=wa -F key=identity
```

The above translates to a list of monitored syscalls, including `rename`, which is hardcoded and added to audit rules even if unavailable on `aarch64`. This patch filters out such syscalls during rule creation.

### DIFF
--- a/lib/libaudit.c
+++ b/lib/libaudit.c
@@ -1530,44 +1530,44 @@ int _audit_parse_syscall(const char *optarg, struct audit_rule_data *rule)
  * on the given architecture. Returns a new string with supported syscalls
  * or NULL on error.
  */
-static char* filter_supported_syscalls(const char* syscalls, int machine) {
-    if (syscalls == NULL) {
-        return NULL;
-    }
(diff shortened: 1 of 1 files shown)


RH-steve-grubb:
filter_supported_syscalls should probably be static unless it can be called from elsewhere. Also, it should probably get the __attr_dealloc_free treatment. (To do that you have to make a separate function declaration.)

Cropi:
> filter_supported_syscalls should probably be static unless it can be called from elsewhere. Also, it should probably get the __attr_dealloc_free treatment. (To do that you have to make a separate function declaration.)

Thanks for the suggestion, I adjusted the patch.

PR #428, opened 2025-01-20
auditd config: remove HALT option from space_left_action

Cropi:
The previously available HALT option, which would cause the audit daemon to shut down the computer system, has been deprecated and should no longer be used. It was determined that halting the system at this stage could lead to unintended consequences and is considered a bad action if selected.

### DIFF
--- a/src/auditd-config.c
+++ b/src/auditd-config.c
@@ -1034,6 +1034,11 @@ static int space_action_parser(const struct nv_pair *nv, int line,
 				if (check_exe_name(nv->option, line))
 					return 1;
 				config->space_left_exe = strdup(nv->option);
+			} else if (failure_actions[i].option == FA_HALT) {
+				audit_msg(LOG_ERR,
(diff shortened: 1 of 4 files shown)


src/auditd-config.c
```diff
@@ -1034,6 +1034,11 @@ static int space_action_parser(const struct nv_pair *nv, int line,
 				if (check_exe_name(nv->option, line))
 					return 1;
 				config->space_left_exe = strdup(nv->option);
+			} else if (failure_actions[i].option == FA_HALT) {
```
Cropi, line 1037: Which one is a better approach?

- Return 1 indicating failure in case HALT is selected for space_left_action (behavior from the patch)
- Log the error via audit_msg and fallback to SYSLOG maybe?

stevegrubb: In all other cases we do both.

stevegrubb:
Thanks!

PR #429, opened 2025-01-20
Add *xattrat syscalls introduced with Linux 6.13

hillu:


stevegrubb:
Looks like this gets every arch except Riscv.

hillu:
That's because the RISCV tables already contain those syscalls.

Cropi:
Yes, I did it on purpose during the review of https://github.com/linux-audit/audit-userspace/pull/73 .

PR #431, opened 2025-01-24
xattrat syscalls: update normalizer and interpreter

Cropi:
Add new table for handling *xattrat at_flags. All of these new syscalls have more than 4 args, so there is a limit on what we can interpret and what we can't.

Regarding testing, I took the master branch and compiled it on fedora-rawhide, where kernel 6.13 is already in place. I invoked all syscalls using custom C programs that directly call the syscalls via the syscall() function.
Part of the patch is a new table, which tracks the so-called "at_flags" arg.

I tested `--format=text,interpret,csv` fields. The `getxattrat` and `listxattrat` syscalls will need additional handling altogether with other existing syscalls used for querying extended attributes. That is gonna be part of another patch.


stevegrubb:
The way that the print_aX functions work is that they check the first letter when there are a number that start with the same letter. If the first letter matches then it does compares until it finds the function. Otherwise it checks the first letter of another group. Eventually it does full string compares because there's no other grouping. This is done to avoid doing a strcmp of every function name in the worst case scenario.

It looks like some are not following this pattern. For example in print_a0, getxattrat is in the 's' group. removexattrat should go in the 'r' group. And with adding listxattrat, an 'l' group can be made.

Also, take a look at print_cap_bitmap. There is some example code in that function that demonstrates how to use stpcpy. It is more efficient to use this than strcat. I know that the code in print_clone_flags needs to be updated to use this technique.

Other than that, looks good.

Cropi:
> The way that the print_aX functions work is that they check the first letter when there are a group of syscalls that start with the same letter. If the first letter matches then it does comparisons until it finds the syscall. Otherwise it checks the first letter of another group. Eventually it does full string compares because there's no other grouping. This is done to avoid doing a strcmp of every function name in the worst case scenario. They are also arranged based on which syscalls are the most frequent to least frequent. The whole point is to find the syscall as fast as possible.
> 
> It looks like some are not following this pattern. For example in print_a0, getxattrat is in the 's' group. removexattrat should go in the 'r' group. And with adding listxattrat, an 'l' group can be made.

I wonder why I put each of them into the wrong place, rookie mistakes. Thanks for pointing it out.

> 
> Also, take a look at print_cap_bitmap. There is some example code in that function that demonstrates how to use stpcpy. It is more efficient to use this than strcat. I know that the code in print_clone_flags needs to be updated to use this technique.

I updated the relevant code, `print_clone_flags` can be updated later if this solution is accepted.

> 
> Other than that, looks good.

stevegrubb:
Yes, let's do those another time. Thanks for updating the PR.

PR #433, opened 2025-02-03
audispd-pconfig.c: don't check unnecessary write permissions

nazarewk:
While trying to use `auditd` with https://github.com/threathunters-io/laurel on NixOS I was getting an error:

	/nix/store/9q60x1fc4yq5av5c5g5h2ccmkprb9h4z-laurel-0.6.4-for-auditd/bin/laurel permissions should be 0750

This is:
- on one side the issues is specific to Nix, where no `/nix/store` file has write access at all (binaries are `0555`)
- on the other side I don't expect `auditd`  to write to that binary file, therefore it should not check for write permissions at all

### DIFF
--- a/audisp/audispd-pconfig.c
+++ b/audisp/audispd-pconfig.c
@@ -506,9 +506,9 @@ static int sanity_check(plugin_conf_t *config, const char *file)
 				 config->path);
 			return 1;
 		}
-		if ((buf.st_mode & (S_IRUSR|S_IWUSR|S_IXUSR|S_IRGRP|S_IXGRP)) !=
-				   (S_IRUSR|S_IWUSR|S_IXUSR|S_IRGRP|S_IXGRP)) {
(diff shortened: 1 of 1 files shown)


Cropi:
Auditd is not supposed to modify the plugin binary. But if the path is not writable by the owner, can you seamlessly update it in case there is a newer version available?

nazarewk:
> But if the path is not writable by the owner, can you seamlessly update it in case there is a newer version available?

Yes, there is a daemon responsible for writing (and building) files to `/nix/store` with pretty much exclusive access to it.

This might be similar for immutable distributions where the root filesystem or package set is never supposed to change.

stevegrubb:
This change is harmless in general. I agree with the check. The main thing is not group writable and not public. I am wondering about the wording. I'll go with it and update the explanation if I think of a better way to explain it.

PR #434, opened 2025-02-04
auditd.cron.5: Describe time-based log rotation setup

Cropi:
Many users frequently ask whether time-based log rotation is possible in `auditd` and how to configure it. The purpose of this change is to provide a general guide on setting up time-based log rotation. Output is a new man page named auditd.cron.

My intention is not to set a strict guideline on how to configure it or what exact values to set in the main config. Feedback is appreciated. Will do some testing later.

### DIFF
--- a/audit.spec
+++ b/audit.spec
@@ -210,6 +210,7 @@ fi
 %attr(644,root,root) %{_mandir}/man8/aulastlog.8.gz
 %attr(644,root,root) %{_mandir}/man8/ausyscall.8.gz
 %attr(644,root,root) %{_mandir}/man5/auditd.conf.5.gz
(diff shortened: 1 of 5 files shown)


stevegrubb:
This looks good. But if the size action is set to ignore, nothing will delete logs. Might need need to touch up the cron script to honor the num_logs setting.

Cropi:
> This looks good. But if the size action is set to ignore, nothing will delete logs. Might need need to touch up the cron script to honor the num_logs setting.

I think it's not needed, because even manual log rotation (via `auditctl --signal rotate`) should respect the num_logs setting. I tried it and I was not able to obtain more log files than the value I've set up.

stevegrubb:
ok, I guess we can merge this. Thanks!

PR #435, opened 2025-02-10
Broadcast a warning if a system halt is possible

Cropi:
The space left action no longer supports HALT. If we reach this stage, ensure that all logged-in users are notified in case `admin_space_left_action` or `disk_space_full_action` is set to halt the system.

### DIFF
--- a/common/common.c
+++ b/common/common.c
@@ -25,6 +25,8 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <stdio.h>
+#include <utmpx.h>
+#include <fcntl.h>
 
 /*
  * This function returns 1 if it is the last record in an event.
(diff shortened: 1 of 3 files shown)


stevegrubb:
OK, I think I'm on board with this. (Was considering if we should make a new target F_HALT_WALL that includes this broadcast while the original does not.) The only thing is the message should say something to the effect of "The audit system is low on disk space and is now halting the system for admin corrective action." It currently tells the why, but not what's about to happen.

Cropi:
Actually, that information is already included in the `next_actions` string.  

1. When the `space_left` threshold is reached, audit will warn about the **HALT** option configured in either `admin_space_left_action` or `disk_full_action`. 
2. When the `admin_space_left` threshold is reached, audit will warn about the **HALT** option in `disk_full_action`.

stevegrubb:
So, it's going to output: "Broadcast message from audit daemon: The audit system is low on disk space for logging. If the disk becomes full, audit will halt."  I feel like it's better to be more direct that it's going down now.

Cropi:
I adjusted the message but there is a reason I did not want to say that explicitly. Since space_left_action no longer includes HALT, corrective actions might intervene, and stating that we're going down could be misleading.

stevegrubb:
Thanks!

PR #437, opened 2025-02-25
Fix compiler warnings

Cropi:
The first two are due to the nature of realloc. It's safer to use a temporary var, so rule can still hold the old ptr in case realloc fails.
The third and fourth are minor issues that crept in during my recent refactoring (sigh).

**1.**
```
libaudit.c: In function 'audit_add_watch_dir':
libaudit.c:805:17: warning: pointer 'rule_19' may be used after 'realloc' [-Wuse-after-free]
  805 |                 free(rule);
      |                 ^~~~~~~~~~
libaudit.c:803:18: note: call to 'realloc' here
  803 |         *rulep = realloc(rule, len + sizeof(*rule));
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

**2.**
```
libaudit.c: In function 'audit_rule_fieldpair_data':
libaudit.c:1839:33: warning: pointer 'rule_335' may be used after 'realloc' [-Wuse-after-free]
 1839 |                                 free(rule);
      |                                 ^~~~~~~~~~
libaudit.c:1837:34: note: call to 'realloc' here
 1837 |                         *rulep = realloc(rule, sizeof(*rule) + rule->buflen);
```

**3.**
```
auditctl.c: In function ‘opt_status’:
auditctl.c:645:13: warning: unused variable ‘retval’ [-Wunused-variable]
  645 |         int retval = args->retval;
      |             ^~~~~~
```

**4.**
```
auditctl.c: In function ‘opt_trim’:
auditctl.c:1201:1: warning: control reaches end of non-void function [-Wreturn-type]
 1201 | }
      | ^
```

### DIFF
--- a/lib/libaudit.c
+++ b/lib/libaudit.c
@@ -789,7 +789,7 @@ int audit_add_watch_dir(int type, struct audit_rule_data **rulep,
 			const char *path)
 {
 	size_t len = strlen(path);
-	struct audit_rule_data *rule = *rulep;
+	struct audit_rule_data *rule = *rulep, *tmp;
 
 	if (rule && rule->field_count) {
 		audit_msg(LOG_ERR, "Rule is not empty");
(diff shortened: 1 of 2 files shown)


stevegrubb:
In opt_trim, there's an assignment to retval and the very next line it's overwritten. Otherwise looks good. (realloc is hard to use correctly.)

Cropi:
Code updated.

stevegrubb:
Thanks!

xuraoqing:
in case 1，why memory that rule pointed to becomes invalid when realloc failed?  if so, use tmp can not solves the leak, i think.

PR #445, opened 2025-04-02
test suite: replace auid=42 with auid=0

Cropi:
Executing make check, the test case expected the system to have user gdm with id of 42, which might not be true in all cases. In case the user was not present, id-to-name translation failed, thus make check exited with error.

### DIFF
--- a/auparse/test/auparse_test.ref
+++ b/auparse/test/auparse_test.ref
@@ -188,7 +188,7 @@ event 4 has 3 records
         uid=0 (root)
         subj=system_u:system_r:init_t:s0 (system_u:system_r:init_t:s0)
         old-auid=4294967295 (unset)
-        auid=42 (gdm)
+        auid=0 (root)
         tty=(none) ((none))
         old-ses=4294967295 (4294967295)
         ses=1 (1)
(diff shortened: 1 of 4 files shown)


sergio-correia:
What happens if gdm is installed and its uid is 42 (or something else), then, after this patch?

Cropi:
The auparse_test program uses pre-generated(I would call  artificial) logs, e.g. `auparse/test/test.log` when testing. Basically, the test parses the files and prints out data about each event.
In case it finds `auid=whatever` field, it will try to translate the ID to name based on what is available on the system.
So if there is a username gdm with id of 42, it will output auid=42(gdm). But if no such user exists on the system, it will output 42(unknown).  In the end, the test will compare expected results auid=42(gdm) with current results auid=42(unknown), so make check will fail.

This is expected because the logs have not been generated on that particular system.
So my suggestion is to rewrite it to root. It will always have an ID of 0.

stevegrubb:
I kind of feel like the test suite should be reworked. Since the logs were captured in 2007, we added the enriched log format. I also don't know how extensive the code coverage is. I don't think there are any tests for the normalizer.

One thing about this change, root is handled differently than any other uid. It does a direct check against 0 and uses root if it's true. Otherwise it does a real lookup. So, the records need to have a few accounts that cause some real lookups in them. We can merge this as is and fix it some other day when the tests are improved.

PR #452, opened 2025-05-21
Remove awk call in condrestart legacy script

pebolle:
This is basically a follow up to yesterday's commit 483b03b88686c9c2088d3117394046d8d6ac3917 ("Remove awk dependency"). That commit was equivalent to a recent change in Fedora's audit package.

So another trivial but BRAVELY UNTESTED pull request.


stevegrubb:
Thanks

PR #453, opened 2025-05-21
Add github actions

Cropi:
Add CI Matrix workflow to test builds on Ubuntu and Fedora with both gcc and clang compilers. Tests are temporarily disabled for Ubuntu builds (see [#445](https://github.com/linux-audit/audit-userspace/pull/445))


stevegrubb:
Go ahead and add it. We can tweak it as it gets used.

Issue #457, opened 2025-06-03
Executing auditd -s enable hangs


juyin:
When the libgcc_s.so.1 library (required when using the pthread_cancel function) is not present in the device environment, the program will hang when executing auditd -s enable. The steps are as follows:
1. Execute /etc/init.d/auditd start to start auditd.
2. Execute auditd -s enable.

The phenomenon is as follows:
![Image](https://github.com/user-attachments/assets/a7b110e8-b14c-45c7-a5bf-e10765e5cdc4)

stevegrubb:
The patch was merged. Closing.

PR #458, opened 2025-06-03
Fix auditd -s enable hang issue (#457)

juyin:
When the environment lacks libgcc_s.so.1, the program will abort when using the pthread_cancel function.

When executing auditd -s enable, the parent process will hang at the become_daemon function waiting for the child process to write to init_pipe[1]. The child process calls shutdown_events at line 972 in auditd.c, which in turn calls pthread_cancel. Due to the absence of libgcc_s.so.1 in the environment, the child process is terminated here, and the parent process continues to wait because it did not receive the write to init_pipe[1] from the child process, causing the program to hang.

Therefore, before calling shutdown_events, the parent process should be notified that it has already failed to ensure the parent process exits promptly.


stevegrubb:
I suppose it doesn't matter to move this up a couple lines. I'lll go ahead and merge it. But, solving problems for broken environments is not a path I want to do down.

PR #467, opened 2025-06-24
auditd: support loading plugin configs from symlinks

LordGrimmauld:
auditd: support loading plugin configs from symlinks

`auditd` previously did not support loading plugin configurations from
symlinked config files. This is problematic on systems such as NixOS,
which constructs basically the entirety of /etc using symlinks.

I considered why symlinks were not supported, and concluded the reason was
simplicity. While having a symlink point to a writable location would be
insecure, a user putting an insecure symlink to trigger this behavior could
also immediately do worse things.

There also were edge cases if the config file is replaced between the file
type check and the actual read. This is because `load_plugin_conf` uses
path based logic to check whether a file is a regular file or not, and then
asses the file path to `load_pconfig`. This means audispd would already
load symlinked configs if a regular file was replaced by a symlink at
precisely the right time in execution.

`load_pconfig` opens the supplied config file path using `open`. Crucially,
it does not set `O_NOFOLLOW`, meaning `load_pconfig` already supports
loading plugin configs from symlinks. The check in `load_pconfig` also
already uses the file-descriptor based `fstat` call, which mitigates the
replacement problems: file descriptors are stable.

This means, to support symlinks, it is sufficient to remove the check for
regular files from `load_plugin_conf`. This does change internal API: It
now is the responsibility of `load_pconfig` to make sure a plugin config
file is a regular file. This API change is purely internal, neither
`load_pconfig` nor `load_plugin_conf` are part of the public headers.

This change has been tested against auditd 4.0.3 and 4.0.5 in a NixOS VM.
The plugin config files af_unix.conf, au-remote.conf, filter.conf, syslog.conf
all successfully loaded through symlink.

### DIFF
--- a/audisp/audispd-pconfig.c
+++ b/audisp/audispd-pconfig.c
@@ -31,6 +31,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <libgen.h>
+#include <limits.h>
 #include "audispd-pconfig.h"
 #include "private.h"
 
@@ -143,12 +144,10 @@ int load_pconfig(plugin_conf_t *config, int dirfd, char *file)
 
 	clear_pconfig(config);
 
-	/* O_NONBLOCK avoids blocking when opening a FIFO file accidentially.
-	 * It does however block if someone symlinks /dev/ttyX into the plugin directory.
+	/* O_PATH avoids blocking, as no read/seek is done.
 	 * We do not pass O_NOFOLLOW, which allows for symlinked configs.
 	 */
-	mode = O_RDONLY | O_NONBLOCK;
-	rc = openat(dirfd, file, mode);
+	rc = openat(dirfd, file, O_PATH);
 	if (rc < 0) {
 		if (errno != ENOENT) {
 			audit_msg(LOG_ERR, "Error opening %s (%s)", file,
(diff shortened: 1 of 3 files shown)


LordGrimmauld:
Okay, `audispd.c` is the only caller of that method, and neither `load_pconfig` nor `load_plugin_conf` are part of public API, so the stability consideration is mostly mute. I will just remove that check completely.

LordGrimmauld:
Alright, with these considerations, its probably better to just remove that check completely and call it good.

stevegrubb:
That test was added to make sure we have a real file. We recently had an issue filed because a fifo was in a directory and opening it deadlocked the program. So, if we need to allow symlinks, we need to check that it is a either a regular file or a symlink that resolves to a regular file to pass that first check.

LordGrimmauld:
Afaik `open` syscall should "just work" for FIFO, unless you set `O_RDWR` (which you don't). So i am a bit confused.... Is there an issue link to read up on that? Or a repro i can test?

LordGrimmauld:
Ah, you need to set `O_NONBLOCK` for fifo to not stall indefinitely.
That said, if there is a block/char device that does not support non-blocking open, the issue would come back.... Hmm...

LordGrimmauld:
I suppose i can do the realpath+stat again (what i had before the force push), but that still allows a potential attacker to "race" and replace a regular file with a blocking char device to infinitely stall the audit daemon.... Not ideal either.

LordGrimmauld:
Hold on, it now complains about world-writable for what worked perfectly fine before, investigating...
The approach should be fine i think, but needs some fixing.

LordGrimmauld:
Okay, now i am happy. no race-y stat/open, fixed the race with the dirfd, and only actually open the file for reading once all the stat succeeds. This *should* be good now.

LordGrimmauld:
Oh, does this need an entry in the changelog? Sorry, there is no contribution guidelines, so i missed that.

LordGrimmauld:
Thanks for getting this merged, and i do appreciate your work! Indeed this feels on the complicated side, i tried to keep it as simple as possible while properly fixing all edge cases (including TOCTOU and questionable symlinks). Chances are, this could be simpler given more assumptions about what is allowed in the config directory, but alas i was trying to write a thorough patch without imposing any additional assumptions.

Now its time to clean this up on our (nixos) side, i am excited to see more security features gaining support. Thanks for your work, not just here but also with `libcap_ng`, these are quite useful tools!

PR #501, opened 2025-09-11
auditd.service: set LogsDirectory and RuntimeDirectory, remove tmpfiles dependency

LordGrimmauld:
setting LogsDirectory and RuntimeDirectory ensures systemd will create these directories ahead of starting the auditd service. It also ensures the auditd service has write permissions, even if someone might add additional hardening options to the systemd service in the future. As a result, there is just no more need for the tmpfiles rules.


LordGrimmauld:
We might also want to set `LogsDirectoryMode` / `RuntimeDirectoryMode` now that i think about this... The tmpfiles rules had `0700` for log dir. I am thinking the runtime dir should probably be `755` or `644`.

PR #515, opened 2026-02-28
Fix NO_CAST.INTEGER_OVERFLOW in libev ev.c array_nextsize

mikhailnov:
Add explicit casts to size_t in arithmetic expressions to prevent potential integer overflow when calculating array sizes.

Overflows are not very realistic, but improve code robustness and make analyzers happy.

Svace reports:
  1) The value of 'elem * ncur' may overflow (ev.c:2259)
  2) The value of 'ncur + elem + (MALLOC_ROUND - 1)' may overflow (ev.c:2262)
  3) The value of 'elem * *cur' may overflow (ev.c:2275)
  (CWE190, CWE197)

### DIFF
--- a/src/libev/ev.c
+++ b/src/libev/ev.c
@@ -2256,10 +2256,10 @@ array_nextsize (int elem, int cur, int cnt)
   while (cnt > ncur);
 
   /* if size is large, round to MALLOC_ROUND - 4 * longs to accommodate malloc overhead */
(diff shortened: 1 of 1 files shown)


stevegrubb:
Commit 676e6c7e adds a note about how libev is modified for audit use. It appears that a 4.34 release is being worked on by upstream. There's still time to get this in so that when it is pulled in to audit, these issues are fixed. Closing this out.

PR #522, opened 2026-04-06
Add versioned symbol support

gordonmessmer:
in order to improve reliable dependency resolution, I would like to add versioned symbols to the libraries that do not include them today.

This change is intended to support Fedora's recommendation that shared libraries provide versioned symbols: https://docs.fedoraproject.org/en-US/packaging-guidelines/C_and_C++/#_versioned_symbols

Please double-check the symbol files. If any symbols have been added very recently, they might not be in the symbol maps I've generated.


stevegrubb:
Thanks for the contribution. That said, we have a policy of never deleting an API unless there is a soname number bump. We do add APIs but that is compatible with old software and new library. This captures some symbols that start with '_'. Those are internal API's that can't be hidden. If this were more automated it would be more attractive.

gordonmessmer:
> We do add APIs but that is compatible with old software and new library.

Right. The general idea here is that adding symbols to a flat namespace that (potentially) grows with each feature release is backward compatible, but it's also difficult to determine at dependency-resolution time whether a binary and shared library are compatible. If, instead, each feature release gets a new symbol set, then that acts as a feature label, and that's something that can easily be used to determine compatibility before installation.

> This captures some symbols that start with '_'. Those are internal API's that can't be hidden.

Can or can't?

If you mean "can", then we could remove all of the "_*" symbols and add a section that hides those symbols:

```
  local:
    _*;
```

> If this were more automated it would be more attractive

I think you could add a step to a pre-release validation... You'd need to build a binary and then:

```
objdump -T $1 | \
  grep -F .text | \
  awk '{if($6=="Base") {print $7;}}' | \
  c++filt | \
  awk '/[() ]/ {print "    \"" $0 "\";";} \
      !/[() ]/ {print "    " $0 ";";}' | \
  sort
```

Anything printed would need to be added in a new symbol set.

gordonmessmer:
Checking in: Please let me know if there are any remaining changes you'd like to see in this PR.

PR #532, opened 2026-07-06
autls: add Phase 0 TLS policy, ACL, and hardening on top of transport layer

sergio-correia:
Layer Phase 0 TLS policy and operational features on top of the transport base in
  upstream/tls:

  - Identity validation, crypto profiles (standard/FIPS/PQC), and
  tls_auth/tls_crypto_profile config options
  - ACL parser with per-identity enable/disable and live SIGHUP reload (no restart
  needed for client revocation)
  - Structured TLS audit records for connection events
  - Non-blocking SSL_write for server ACKs with proper WANT_READ/WANT_WRITE
  differentiation
  - SIGCHLD signal safety (SA_RESTART + errno preservation)
  - Man pages and sample configs for the PSK tech preview
  - Unit tests for identity validation, profiles, ACL, config parsing, and sanity checks
  - Build system: explicit link dependencies, OPENSSL_LIBS override support

### DIFF
--- a/audisp/plugins/remote/audisp-remote.c
+++ b/audisp/plugins/remote/audisp-remote.c
@@ -249,8 +249,10 @@ static void user2_handler( int sig )
  */
 static void child_handler(int sig)
 {
+	int saved_errno = errno;
 	while (waitpid(-1, NULL, WNOHANG) > 0)
 		; /* empty */
+	errno = saved_errno;
 }
 
 /*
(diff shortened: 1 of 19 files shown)


stevegrubb:
This is a very good chunk of what's needed minimally. There are 5 things that are not quite right. Given how much this accomplishes, I am inclined to take it "as is" and make the small changes. Completing phase 0 would be enough to merge into the master branch and calling it generally available. Thanks.

PR #534, opened 2026-07-07
Fix supported architectures claim

Arnie97:
MIPS support was never merged upstream (#115, #198).

Not to be confused with RISC-V, which is supported now (#73).


stevegrubb:
Thanks

PR #537, opened 2026-07-15
Enable TLS in CI and remove unsupported certificate-mode references

sergio-correia:
- Add --enable-tls and OpenSSL dev packages to the CI build matrix so TLS code is compiled and tested on both GCC/Clang across Ubuntu and Fedora.
- Remove certificate-mode options (tls_cert_file, tls_key_file, tls_ca_file, tls_client_auth) from man pages and default config files, and simplify sanity_check() validation to PSK-only logic — only PSK authentication is functional.


stevegrubb:
ok, looks good. Thanks!

PR #538, opened 2026-07-15
Use _exit in forked children after exec failure

sergio-correia:
Forked children that called exit() after a failed execve() would run the parent's atexit handlers.  In auditd the clean_exit() handler calls audit_set_pid(fd, 0), which deregisters the daemon from the kernel, and unlinks the PID file.  When the real parent keeps running it silently stops receiving audit events.

The same class of problem was fixed for the sendmail helper in 7965bc1a ("Use _exit after sendmail exec failure").  Apply the identical fix to every remaining site:

- common/common.c       change_runlevel() child
- audisp/plugins/ids/reactions.c  safe_exec() child
- audisp/plugins/remote/audisp-remote.c  safe_exec() child
- audisp/plugins/filter/audisp-filter.c  execve() child

### DIFF
--- a/audisp/plugins/filter/audisp-filter.c
+++ b/audisp/plugins/filter/audisp-filter.c
@@ -514,13 +514,13 @@ int main(int argc, const char* argv[])
 		    config.binary_args[0] == NULL) {
 			syslog(LOG_ERR,
 			       "audisp-filter: missing child command");
-			exit(1);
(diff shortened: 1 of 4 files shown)


stevegrubb:
The filter plugin needs the same treatment. Otherwise looks good.

sergio-correia:
> The filter plugin needs the same treatment. Otherwise looks good.

Updated, thanks for the review.

stevegrubb:
Thanks!

PR #539, opened 2026-07-17
tls: close certificate-authentication fallback and enforce PSK identity

sergio-correia:
Remove the certificate-mode authentication code that was never shipped but left a fallback path where a non-PSK client could complete a TLS 1.3 handshake when both PSK callbacks and certificate credentials were loaded into the same SSL_CTX. Add a post-handshake check that rejects any client without an accepted PSK identity, and fail the PSK callback immediately if identity storage fails.


stevegrubb:
ok, looks good. Thanks!

PR #540, opened 2026-07-24
Fix comm truncation, manpage check, completions and tmpfiles for /run/audit

sergio-correia:
Here are a couple of small fixes for issues identified when packaging audit 4.2:
- libaudit: Truncate overlong comm values instead of rejecting them, preventing audit rule failures for processes with long command names
- docs: Skip .so redirect stubs in the manpage completeness check so they don't cause false failures
- init.d: Create /run/audit via tmpfiles.d instead of relying on the daemon to create it at startup
-  init.d: Install bash completions under per-command names (auditctl, ausearch, aureport, augenrules) so they auto-load; previously installed as audit.bash_completion which bash-completion never discovered

### DIFF
--- a/docs/audit_log_user_comm_message.3
+++ b/docs/audit_log_user_comm_message.3
@@ -33,7 +33,8 @@ The kernel task command name is limited to 15 bytes because
 .B TASK_COMM_LEN
 is 16 bytes including the terminating null byte. A non-NULL
 .I comm
-value longer than 15 bytes is malformed and is rejected.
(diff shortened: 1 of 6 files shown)


stevegrubb:
Actually, this needs a followup for "install bash completions under per-command names" to fix the spec file for the change.

sergio-correia:
Right, I forgot there was a spec file also in this repo. Let me submit a quick PR

PR #541, opened 2026-07-24
audit.spec: update bash completion filenames

sergio-correia:
Reflect the per-command completion names from the Makefile.am change in commit 3e66e559.


stevegrubb:
Thanks

PR #544, opened 2026-08-10
auditd: fix V0 dispatch truncating last 16 bytes of events

Cropi:
When log_format=RAW and any audisp plugin (e.g. audisp-syslog) is active,
the last 16 bytes of every kernel audit event are silently dropped before
reaching the plugin pipe. A USER record ending in `terminal=? res=success'`
shows up in syslog as `termina` with the rest gone.

**Root cause**

Commit ba8ba4f2 introduced a branch for embedded netlink replies
(`rep->nlh == &rep->msg.nlh`) that subtracted `NLMSG_HDRLEN` (16) from
`nlmsg_len` before copying the payload, assuming the kernel follows the
standard netlink convention where `nlmsg_len = header + payload`.

The kernel audit subsystem doesn't follow that convention.
In `kernel/audit.c`, `__audit_log_end()` sets:

    nlh->nlmsg_len = skb->len - NLMSG_HDRLEN;

So `nlmsg_len` is already the payload size. Subtracting `NLMSG_HDRLEN` again
cuts 16 real bytes off every kernel audit event.

The kernel documents this explicitly in `kauditd_send_multicast_skb()`:

> "non-standard mods are made to the skb by the original kaudit unicast
> socket send routine. The existing auditd daemon assumes this breakage.
> Fixing this would require co-ordinating a change in the established
> protocol between the kaudit kernel subsystem and the auditd userspace
> code."

The multicast path fixes `nlmsg_len` back to `skb->len` for non-auditd
readers, but the unicast path that auditd reads doesn't.

**Fix**

Remove the `rep->nlh == &rep->msg.nlh` branch distinction. The kernel always
sets `nlmsg_len` to the payload size, so use it directly with no subtraction.
Keep the bounds check against the buffer size.

**Tests**

- `test_netlink_payload_length`: updated to use payload-only `nlmsg_len`,
  matching how the kernel actually sets it
- `test_invalid_netlink_length`: removed the now-impossible low-bound case
  (`< NLMSG_HDRLEN`)
- `test_realistic_user_event`: new -- full USER record payload round-trip
  through V0 dispatch
- `test_v2_protocol_uses_rep_len`: new -- VER2 path uses `rep->len`, not
  `nlmsg_len`; confirms no regression there

**How to reproduce**

1. Set `log_format=RAW` and enable audisp-syslog
2. Generate events: `auditctl -m "test message"`
3. Compare `audit.log` vs syslog output -- syslog entries will be exactly
   16 bytes shorter every time

### DIFF
--- a/src/auditd-dispatch.c
+++ b/src/auditd-dispatch.c
@@ -75,24 +75,14 @@ int dispatch_event(const struct audit_reply *rep, int protocol_ver)
 	e->hdr.type = rep->type;
 
 	if (protocol_ver == AUDISP_PROTOCOL_VER) {
-		if (rep->nlh == &rep->msg.nlh) {
-			if (rep->msg.nlh.nlmsg_len < NLMSG_HDRLEN ||
-			    rep->msg.nlh.nlmsg_len >
-					NLMSG_LENGTH(sizeof(e->data))) {
-				free(e);
-				return -1;
-			}
-
-			/* audit_get_reply marks embedded netlink replies this
-			 * way. Netlink length includes its header, while
-			 * event_t contains only payload. Local V0/legacy
-			 * events retain a payload length here. */
-			e->hdr.size = rep->msg.nlh.nlmsg_len - NLMSG_HDRLEN;
-		} else if (rep->msg.nlh.nlmsg_len > sizeof(e->data)) {
+		/* The kernel audit subsystem sets nlmsg_len to the payload
+		 * size only (skb->len - NLMSG_HDRLEN), not the total
+		 * message length. */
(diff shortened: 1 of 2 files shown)


sergio-correia (approved):
Nice catch! The kernel audit subsystem's non-standard `nlmsg_len` behavior makes this code deceptively tricky, as it is easy to assume standard netlink conventions apply, but
`__audit_log_end()` deliberately sets `nlmsg_len = skb->len - NLMSG_HDRLEN` on the unicast path. The old subtraction was a double-subtract bug.

stevegrubb:
Thanks.

PR #545, opened 2026-08-12
Add Landlock audit types

Skyb0rg007:
Audit messages produced by Landlock were written as `type=UNKNOWN[1423]` in `/var/log/audit/audit.log`, while the [Landlock docs](https://docs.kernel.org/admin-guide/LSM/landlock.html) uses the `LANDLOCK_ACCESS` and `LANDLOCK_DOMAIN` names.

There is also an issue with how `mode=` is handled -- Landlock outputs `mode=enforcing`, while interpret.c treats it like a file permission and tries to parse it into a string.
I didn't attempt to fix the issue as it's not relevant at the moment (`enforcing` is the only Landlock mode).

### DIFF
--- a/lib/audit-records.h
+++ b/lib/audit-records.h
@@ -104,6 +104,10 @@ extern "C" {
 #define AUDIT_LAST_EVENT        1399
 
 #define AUDIT_FIRST_SELINUX     1400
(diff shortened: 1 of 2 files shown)


stevegrubb:
Thanks.

PR #547, opened 2026-08-12
init.d: Install tmpfiles config file under more relaxed perms

krnowak:
Contents of this file are not a secret that should not be read by other users/tools.

In Flatcar are doing some postprocessing of tmpfiles config files when building the images and this one file failed to be read because of permission being denied.


stevegrubb:
Thanks.

PR #551, opened 2026-09-14
audisp-syslog: include <unistd.h> for getppid()

itxaka-agent:
### Problem

`audisp/plugins/syslog/audisp-syslog.c` calls `getppid()` from its `SIGTERM` handler but never includes `<unistd.h>`.

On glibc the prototype leaks in via another header the plugin already pulls in, so the build stays quiet. On musl there is no such leak and the compile fails:

```
audisp-syslog.c: In function 'term_handler':
audisp-syslog.c: error: implicit declaration of function 'getppid'
```

Every sibling audisp plugin (`af_unix`, `filter`, `ids`, `remote`, `statsd`, `zos-remote`) already includes `<unistd.h>` for the same reason — this just matches them.

### Downstream context

Hit while building `audit-userspace` 4.2.1 for [Hadron Linux](https://github.com/kairos-io/hadron) (musl-based minimal distro). Carrying the patch downstream, but the fix belongs upstream so every musl consumer stops hitting it.


Itxaka:
PR was automated with agent but fix was done manually by me a human, as commented, while adding autdit to hadron I hit this issue.

stevegrubb:
Looks good. Thanks for the patch.

Itxaka:
> Looks good. Thanks for the patch.

that was fast, thank you!

Issue #552, opened 2026-09-16
Potential integer overflow at src/libev/ev.c


samitistfobos-dev:
Potential problem
In `array_nextsize` (src/libev/ev.c, 2249), the size of the allocated array is calculated using `int` operands. The expression `elem * ncur` may exceed the range of `int` for sufficiently large values of `ncur`, causing signed integer overflow.
The same multiplication is subsequently performed by `ncur *= elem`, so changing only the comparison would leave another potentially overflowing operation.

Possible solution
It is suggested to use a separate `long` variable for calculations involving the array size in bytes. The multiplication can then be performed as `(long)elem * ncur`, and subsequent rounding calculations can also use this wider type before converting the resulting element count back to `int`.
This prevents signed integer overflow in the byte-size calculations while preserving `ncur` as the number of array elements.

Found by Linux Verification Center (portal.linuxtesting.ru) with SVACE.

Author A. Burlakov.

stevegrubb:
Libev is a "vendored" library. Fixes for it should be sent to the mail list on this page: https://software.schmorp.de/pkg/libev.html. I have some small patches applied to it that upstream won't take. But I don't want to carry any patches because I will lose them on the next upgrade of libev.

samitistfobos-dev:
I sent an email, thank you. I'll write here when I get a reply.

PR #553, opened 2026-09-16
Fix: NO_CAST.INTEGER_OVERFLOW

samitistfobos-dev:
Problem:
The value of an arithmetic expression 'elem * ncur' is a subject to overflow because its operands are not cast to a larger data type before performing arithmetic

Solution: Use a long variable for byte-size calculations in array_nextsize instead of performing the multiplication using int values. This avoids signed integer overflow in both the size comparison and the subsequent size calculation.
Signed-off-by: a.burlakov@fobos-nt.ru
Signed-off-by: dinchik@altlinux.org


samitistfobos-dev:
Patch for https://github.com/linux-audit/audit-userspace/issues/552

